PR #4, opened 2017-06-06
RAINCATCH-978 - demo app base

wtrocki:
## Motivation

Create typescript version of the cloud demo app

## Progress
- [x] Typescript express sample
- [x] Example route
- [x] Lerna runtime
- [x] Documentation


nialldonnellyfh (commented):
It seems as if this app is half-typescript and half javascript.

I think we should start with typescript if this is the demo app that we are going to work with.

wtrocki:
@nialldonnellyfh Can you reference parts that aren't in typescript.
I think I have covered everything we needed.

demo/server/src/app.ts
```diff
@@ -0,0 +1,55 @@
...
+app.use(bodyParser.json());
+app.use(bodyParser.urlencoded({ extended: false }));
+app.use(cookieParser());
+app.use(express.static(path.join(__dirname, 'public')));
+app.use(favicon(path.join(__dirname, 'public', 'favicon.ico')))
+
+app.use('/', index);
+
+//catch 404 and forward to error handler
+app.use((req: express.Request, res: express.Response, next) => {
```
paolobueno: I think the compiler can infer these types for `req, res, next`, we don't even these annotations :smile:

wtrocki: Assumed the same, but I got tslint issues on this :] 
I will try again.

demo/server/src/app.ts
```diff
@@ -0,0 +1,55 @@
...
+  err.status = 404;
+  next(err);
+});
+
+//error handlers
+
+//development error handler
+//will print stacktrace
+if (process.env.NODE_ENV === 'development') {
+  app.use((err: any, req: express.Request, res: express.Response) => {
```
paolobueno: For error handlers, though, we need a 4-arity function. Ignore unused param linting error if there's one.

If we declare a `express.ErrorRequestHandler` variable it's easier to set it up with the if

paolobueno: example from earlier: https://github.com/feedhenry-raincatcher/raincatcher-typescript/blob/master/apps/cloud/src/app.ts#L54-L73

demo/server/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+ * Module dependencies.
+*/
+import app from './app';
+import * as http from 'http';
+
+/**
+ * Get port from environment and store in Express.
+ */
+const port = normalizePort(process.env.PORT || 3000);
+app.set('port',port);
```
paolobueno: missing space after comma, is this not a lint error? just curious

demo/server/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (error.syscall != 'listen') {
+    throw error;
+  }
+  
+  var bind = typeof port === 'string' ? 'Pipe ' + port : 'Port ' + port;
+  
+  //handle specific listen errors with friendly messages
+  switch(error.code) {
+    case 'EACCES':
+      console.error(bind + 'requires elevated privileges');
```
paolobueno: I'd rather have template strings instead of concat if possible. i.e. `${bind} requires elevated privileges`
This concat for instance is missing the space before `requires`, and that's a kind of error that template strings solve nicely.

Also we'll need to replace `console.error` with the logger module later

demo/server/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+  
+/**
+ * Event listener for HTTP server "listening" event.
+ */
+function onListening() {
+  var addr = server.address();
+  var bind = typeof addr === 'string' ? 'pipe ' + addr : 'port ' + addr.port;
+  console.log('Listening on ' + bind);
+}
```
paolobueno: Missing newline at the end of the file should also be a linting error for now

demo/server/test/index.ts
```diff
@@ -0,0 +1,11 @@
+/// <reference types="mocha" />
+
+function baseSuite() {
+  describe('Base', function() {
+    describe('#foo()', function() {
+      it('should receive a string param', function() {
+      });
+    });
+  });
+}
```
paolobueno: need to update unit tests or just remove them from this package

demo/server/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@raincatcher/demo-server",
+  "version": "1.0.0",
+  "description": "Raincatcher demo application server using express",
+  "types": "src/index.ts",
+  "author": "feedhenry-raincatcher@redhat.com",
+  "license": "Apache-2.0",
+  "scripts": {
+    "clean": "del src/**/*.js src/**/*.map test/**/*.js test/**/*.map",
+    "cleanInstall": "lerna exec -- npm install --ignore-scripts",
```
paolobueno: this one would be for the root /package.json, so it runs on all packages

demo/server/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (error.syscall != 'listen') {
+    throw error;
+  }
+
+  var bind = typeof port === 'string' ? 'Pipe ' + port : 'Port ' + port;
+
+  //handle specific listen errors with friendly messages
+  switch (error.code) {
+    case 'EACCES':
+      console.error(bind + 'requires elevated privileges');
```
paolobueno: missing space here

paolobueno (approved):

README.md
```diff
@@ -118,3 +118,11 @@ We recommend users to write their code in TypeScript, especially if their new Ra
 ![JS completion through editor support](img/js-completion.png)
 
 See the example on [examples/js]() for more information also for reusing the unit test suites from JavaScript code.
+
+### Repository commands
+
+ `npm run test` - run unit tests
```
JameelB, line 124: maybe have these commands in individual lines? At the moment, they're in one paragraph and may be a little hard to read.

demo/server/src/index.ts
JameelB: I don't see that specified under the rules in the tslint config file? Is that rule just given?

demo/server/src/app.ts
```diff
@@ -0,0 +1,53 @@
...
+} else {
+  errHandler = (err: any, req: express.Request, res: express.Response, next: () => void) => {
+    res.status(err.status || 500);
+    res.render('error', {
+      title: 'error',
+      message: err.message,
+      error: {}
+    });
+  };
+}
```
witmicko: dry it up a bit with
```
errHandler = (err: any, req: express.Request, res: express.Response, next: () => void) => {
      const idDev: boolean = process.env.NODE_ENV === 'development';
      res.status(err.status || 500);
      res.render('error', {
          title: 'error',
          message: err.message,
          error: isDev ? err : {}
      });
    };
```

?

demo/server/src/app.ts
```diff
@@ -0,0 +1,53 @@
...
+import * as path from 'path';
+import * as logger from 'morgan';
+import * as cookieParser from 'cookie-parser';
+import * as bodyParser from 'body-parser';
+import index from './routes/index';
+import * as favicon from 'serve-favicon';
+
+const app: express.Express = express();
+
+app.use(logger('dev'));
```
nialldonnellyfh: `'dev'` should be a constant / config?

witmicko (approved):
one comment, lgtm otherwise

demo/server/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Module dependencies.
+*/
+import app from './app';
+import * as http from 'http';
+
+/**
+ * Get port from environment and store in Express.
+ */
+const port = normalizePort(process.env.PORT || 3000);
+app.set('port', port);
```
nialldonnellyfh: `'port'` should be a constant / config?

demo/server/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Listen on provided port,on all network interfaces.
+ */
+server.listen(port, onListening);
+server.on('error', onError);
+
+/**
+ * Normalize a port into a number,string,or false.
+ */
+function normalizePort(val: any): number | string | boolean {
```
nialldonnellyfh: Should we enforce `val` to be a number?

demo/server/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    return port;
+  }
+
+  return false;
+}
+
+/**
+ * Event listener for HTTP server "error" event.
+ */
+function onError(error: any) {
```
nialldonnellyfh, line 28: Can we use `@types/node` to identify the error type coming back?

E.g. http://mherman.org/blog/2016/11/05/developing-a-restful-api-with-node-and-typescript/#.WTkimmSGNTY

demo/server/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+  if (error.syscall != 'listen') {
+    throw error;
+  }
+
+  var bind = typeof port === 'string' ? 'Pipe ' + port : 'Port ' + port;
+  //handle specific listen errors with friendly messages
+  switch (error.code) {
+    case 'EACCES':
```
nialldonnellyfh: Can probably use the node typings to get the error types

nialldonnellyfh (commented):
@wtrocki Added some comments.

demo/server/src/index.ts
witmicko: @JameelB yes, we extend on these recommended rules https://github.com/palantir/tslint/blob/master/src/configs/recommended.ts

demo/server/src/index.ts
witmicko: +1 `(val: number | string)`

demo/server/src/index.ts
wtrocki: Sadly there is no standard in that name as node in different versions using integers etc as types and actual error comes from low level library.
Done refactor for this anyway. It's more clear now.

demo/server/Gruntfile.js
```diff
@@ -0,0 +1,34 @@
...
+                files: ['src/**/*.ts'],
+                tasks: ['exec:run'],
+                options: {
+                    livereload: true,
+                    atBegin: true
+                },
+            },
+        },
+        exec: {
+            'run': 'npm run start"',
```
JameelB: is that extra `"` supposed to be there?

demo/server/src/util/config.ts
```diff
@@ -0,0 +1,43 @@
+/** 
+* Interface for fetching application configuration. 
+* 
+* @type T specifies interface used to wrap configuration
+*/
+export interface Config<T> {
```
wtrocki, line 6: @paolobueno We may extract this to some common place in future.

wtrocki:
I merged that because additional commits were addressing concerns but ideally this should be reviewed again.

PR #7, opened 2017-06-07
RAINCATCH-911 refactor

witmicko:
bit of a cleanup
@wtrocki

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ yarn-debug.log*
 yarn-error.log*
 lerna-debug.log*
 
+package-lock.json
 # Runtime data
 pids
 *.pid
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -7,6 +7,7 @@ yarn-debug.log*
 yarn-error.log*
 lerna-debug.log*
 
+package-lock.json
```
paolobueno, line 10: this is supposed to go in version control if using npm>=5

don't think there's any harm to it

witmicko: aye, wasnt sure if we want them in there at the moment, will add them so.

wtrocki:
Integrated manually due to conflicts.

PR #8, opened 2017-06-08
[WIP] RAINCATCH-956 logger interface

witmicko:
## Motivation

Create typescript interface for logger utility
JIRA [RAINCATCH-956](https://issues.jboss.org/browse/RAINCATCH-956)

## Progress

- [ ] Interface
- [ ] Unit tests
- [ ] Sample Implementation
- [ ] Functional tests
- [ ] Documentation
- [ ] Publishing

temp:
logged out like this:
```bash
 ✘  ~/raincatcher/raincatcher-core/cloud/logger-debug   RAINCATCH-956_logger_interface ●✚  mocha


  RaincatcherBase
    #customFunction
  NoopRaincatcherBase:debug: asdasd +0ms
      ✓ should add the supplied prefix
    Base
      #foo()
  base siute:error: hello +2ms
        ✓ should receive a string param


  2 passing (10ms)
```

### DIFF
--- a/cloud/logger-debug/src/index.ts
+++ b/cloud/logger-debug/src/index.ts
@@ -1,0 +1,63 @@
+import * as debug from 'debug';
+import Logger, {LOG_LEVEL} from './Logger';
+
+/**
+ * Raincatcher logger implementation, can take a different logger at construction time.
+ */
+export default class LoggerImpl implements Logger {
+
+  public static getInstance(): LoggerImpl {
+    return LoggerImpl.instance;
+  }
+
+  private static instance: LoggerImpl = new LoggerImpl();
+  private namespace: string;
+  private logger: any;
+
+  constructor() {
+    if (LoggerImpl.instance) {
+      throw new Error('Error: Instantiation failed: Use LoggerImpl.getInstance() instead of new().');
+    } else {
+      LoggerImpl.instance = this;
+    }
+  }
+
+  public setNamespace( namespace: string) {
+    this.namespace = namespace;
+    this.logger = debug(`${namespace}:error:`);
+  }
+
+  public debug(message: string): void {
+    this.logger(`debug:${message}`);
+  }
+
+  public error(message: string): void {
+    this.logger(`error:${message}`);
+  }
+
+  public info(message: string): void {
+    this.logger(`info:${message}`);
+  }
+
+  public warn(message: string): void {
+    this.logger(`warning:${message}`);
+  }
+
+  public log(level: LOG_LEVEL, message: string) {
+    switch (+level) {
+      case LOG_LEVEL.DEBUG:
+        this.debug(message);
+        break;
+      case LOG_LEVEL.ERROR:
+        this.error(message);
+        break;
+      case LOG_LEVEL.INFO:
+        this.info(message);
+        break;
+      case LOG_LEVEL.WARN:
+        this.warn(message);
+        break;
+      default:
+    }
+  }
+}
--- a/cloud/logger-debug/.gitignore
+++ b/cloud/logger-debug/.gitignore
@@ -1,0 +1,5 @@
+# generated code
+src/**/*.js
+src/**/*.map
+test/**/*.js
+test/**/*.map
--- a/cloud/logger-debug/README.md
+++ b/cloud/logger-debug/README.md
@@ -1,0 +1,13 @@
+# RainCatcher Logger
+
+
+Based of example base module located in `/raincatcher-core/examples/base`
+
+//TODO
+Raincatcher Logging standards
+
+//TODO
+module description
+
+//TODO
+module usage
--- a/cloud/logger-debug/index.js
+++ b/cloud/logger-debug/index.js
@@ -1,0 +1,10 @@
+/**
+ * This file is only a proxy for src/index, so it can be required by either
+ * JavaScript or TypeScript code.
+ *
+ * If running code through `ts-node` it'll require src/index.ts. If running from
+ * a nodejs environment it'll require src/index.js,
+ * which is generated via compilation.
+ */
+
+module.exports = require('./src/');
--- a/cloud/logger-debug/package.json
+++ b/cloud/logger-debug/package.json
@@ -1,0 +1,43 @@
+{
+  "name": "@raincatcher/logger",
+  "version": "0.0.1",
+  "description": "RainCatcher logger utility",
+  "types": "src/index.ts",
+  "author": "feedhenry-raincatcher@redhat.com",
+  "license": "Apache-2.0",
+  "scripts": {
+    "clean": "del src/**/*.js src/**/*.map test/**/*.js test/**/*.map",
+    "build": "tsc",
+    "test": "nyc mocha",
+    "prepublish": "npm run clean && npm run build"
+  },
+  "nyc": {
+    "include": [
+      "src/**/*.ts"
+    ],
+    "extension": [
+      ".ts"
+    ],
+    "require": [
+      "ts-node/register"
+    ],
+    "check-coverage": true,
+    "lines": 75,
+    "functions": 100,
+    "branches": 80
+  },
+  "devDependencies": {
+    "@types/mocha": "^2.2.41",
+    "@types/node": "^7.0.29",
+    "@types/debug": "0.0.29",
+    "del-cli": "^1.0.0",
+    "mocha": "^3.4.2",
+    "nyc": "^11.0.1",
+    "source-map-support": "^0.4.15",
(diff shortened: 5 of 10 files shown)


examples/base/package-lock.json
```diff
@@ -0,0 +1,1691 @@
+{
```
wtrocki: Please do not include shrinkwrap in project. Product will get those in productization process.
See productization ticket for more details.

lerna.json
```diff
@@ -4,7 +4,8 @@
     "cloud/*",
     "client/*",
     "examples/*",
-    "demo/*"
+    "demo/*",
+    "utilities/*"
```
wtrocki: IMHO let's put that initially to cloud. 
As we categorize our modules for platform (server,client).
Having utilities will require some changes to documentation diagrams and tickets. 

Client side can be done as separate PR as it will have much differences (will need to be adapted to differnt js frameworks like react native)

utilities/logger/.gitignore
```diff
@@ -0,0 +1,5 @@
+# generated code
+src/*.js
```
wtrocki: proper format is src/**/*.js

wtrocki: Done changes in separate PR. Going to merge soon

utilities/logger/src/LoggerInterface.ts
```diff
@@ -0,0 +1,9 @@
+interface LoggerInterface {
+  init(logger: any, options: any): void;
```
wtrocki: If we research interfaces for other loggers we can notice that interfaces always have 2 levels of usage.

`Loogger.log(LoggerConst.Level_DEBUG,"Message")`

From this we can build actual helper methods for most common levels.
You can take look on: https://www.slf4j.org, which is exactly something we want to build here.

utilities/logger/src/LoggerInterface.ts
```diff
@@ -0,0 +1,9 @@
...
+  debug(message: string, data: any, source: string, tags: string[]): void;
+  error(message: string, data: any, source: string, tags: string[]): void;
+  info(message: string, data: any, source: string, tags: string[]): void;
+  warn(message: string, data: any, source: string, tags: string[]): void;
```
wtrocki: Really strange format to me. I will start with functional tests first by drafting example use cases. 
Once you will do it you will see actual problems with current interface :)
Do not want to spoil that :D

utilities/logger/src/client.ts
```diff
@@ -0,0 +1,28 @@
+import LoggerInterface from './LoggerInterface';
+
+/**
+ * Client logger implementation
+ */
+
+export class Client implements LoggerInterface {
```
wtrocki: I do not think that code has any value.

utilities/logger/src/LoggerInterface.ts
```diff
@@ -0,0 +1,9 @@
+interface LoggerInterface {
```
wtrocki: In the rules you were drafting we were saying that interface names should not be used: `ILogger` or `LoggerInterface` etc. let's keep this standard.

utilities/logger/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export {Core} from './core';
+export {Server} from './server';
```
wtrocki: Client and server will have different requirements for logging and will require separate modules.
We simply cannot have them together. 
Let's remove this and follow our pattern in README: expose implementation and named interface.

utilities/logger/.gitignore
witmicko: copied from base example, will fix in both.

utilities/logger/src/LoggerInterface.ts
witmicko: agree, I ll make data source and tags optional, however, they are encouraged as per https://issues.jboss.org/browse/RAINCATCH-989

utilities/logger/src/client.ts
witmicko: do you mean that the implementation has no value or this code has no value? its only a placeholder really implementing interface.

utilities/logger/src/index.ts
witmicko: +1, assumed it would be factored out into separate module at a later stage as we need three different implementations.
I ll drop client and server and implement core only for now

utilities/logger/src/client.ts
wtrocki: Both.. Let's bring logger to the party and implement that.

utilities/logger/src/index.ts
wtrocki: There will be no more work around login for this so implementation should be complete. We may have tickets in case of some issues, but try to cover and write framework style logger. You can check existing logging solutions or how other frameworks doing loggers: 
see feathers.js for example

cloud/datasync/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@raincatcher/datasync-cloud",
```
wtrocki: :)

cloud/datasync/src/index.ts
```diff
@@ -0,0 +1,145 @@
+import * as Promise from 'bluebird';
+import * as Logger from 'bunyan';
+import * as Express from 'express';
+import SyncGlobalOptions from './syncOptions';
```
wtrocki: Always use base example as base :)

wtrocki: oh.. my bad. It looks like some unmerged commits appeared in this PR. Let's drop them!

cloud/logger/src/Logger.ts
```diff
@@ -0,0 +1,8 @@
+interface Logger {
+  debug(message: string, data?: any, source?: string, tags?: string[]): void;
```
wtrocki: what is source, tags ? I think that source is actually determined in logger constructor.
I would like to see documentation for this to fully know what we talking about here.

Can we add generic method as well `log(level: ErrorLevel, message..`

cloud/logger/src/index.ts
```diff
@@ -0,0 +1,50 @@
+import Logger from './Logger';
+
+
+export default class LoggerImpl implements Logger {
```
wtrocki: Missing info what's the purpose of this class. Just description why we need this instead of explicit BunyanLogger.

cloud/logger/src/Logger.ts
witmicko: the source is the file in which it logs - it's very helpful in browserified apps.
will add generic log

wtrocki: Browserify happening on client. Client and cloud will have different interface.

witmicko: disagree, I think we should have one interface. Browserify use case was just an example, source tag is recommended practice outlined in engineering standards linked here https://issues.jboss.org/browse/RAINCATCH-989

cloud/datasync/src/index.ts
paolobueno: probably need a rebase so 01636ff branches off of master instead of the sync branch

cloud/logger/src/index.ts
```diff
@@ -0,0 +1,77 @@
+import Logger from './Logger';
+import {LOG_LEVEL} from './Logger';
+
+/**
+ * Raincatcher logger implementation, can take a different logger at construction time.
+ */
+export default class LoggerImpl implements Logger {
+  private logger: any;
```
paolobueno: -1 from me, I'd rather have our logger module implement the interface on top of a specific lib like bunyan/debug and leave others for upstream contribution through something like `@raincatcher/logger-debug`.

The semantics for most of the libs around these parameters are very different, so just forwarding the arguments likely won't work.

i.e. for bunyan we have to add the extra data to the object we pass to it, and for debug the tags would be prefixes and so on.

paolobueno: What I mean is that while this looks like a nice generic implementation, it doesn't look like it'll work properly with any library, not even the native `console`, so we probably can't give up having specific implementations instead.

cloud/logger-debug/src/index.ts
```diff
@@ -0,0 +1,70 @@
+import * as debug from 'debug';
+import Logger from './Logger';
+import {LOG_LEVEL} from './Logger';
```
wtrocki: you can merge this with line 2

wtrocki:
**Thanks for contribution!** 
Basing on PR description and state we think that PR is outdated and it's going to be closed. 
Feel free to create another PR or issue if needed.

PR #9, opened 2017-06-12
[RAINCATCH-869] Implement Basic Security Interface for Passport Authentication

JameelB:
## What 
Create an interface for Passport.js authentication functionality.

## Why 
In order to provide passport authentication. The same interface style as Keycloak needs to be followed in order to swap out authentication modules easily. 

**Status: In Progress**

- [x] Add a package with a base skeleton for passport-auth
- [x] Test passport authentication with the demo server routes
- [x] Create a passport-auth module under cloud/
- [x] Create an interface for passport authentication
- [x] Implement the defined interface
- [x] Create integration demo within the passport-auth module
- [x] Add tests
- [x] Add documentation

Pinging @wtrocki


demo/server/src/app.ts
```diff
@@ -1,33 +1,112 @@
...
 
+//passport config
+authService.setSerializeFunction(function(user: User, done: Function) {
+  done(null, user.id);
+});
+
+authService.setDeserializeFunction(function(id: string, done: Function) {
+  users.forEach(function(user) {
+    if(user.id === id) {
+      let userObj = {
```
tomjackman: Is this necessary to have all user fields in the `req.user` or would the user's ID suffice? Just thinking it might render the proposed `getProfileData()` method in the auth interface to become a little redundant, which is there so we can copy Keycloak's approach.

demo/server/src/auth/local.ts
```diff
@@ -0,0 +1,10 @@
+import * as express from 'express';
+
+export function loginRequired(req: express.Request, res: express.Response, next: Function) {
```
tomjackman: Is this method going to be the same as the higher level `keycloak.protect()` to protect routes and check auth/roles, or is it going to be similar to the `login-required` option in the keycloak init method. Just the naming is causing a bit of confusion.

packages/passport-auth/src/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  public setDeserializeFunction(deserialize: Function) {
+    //Retrieves user object from the source using the id as an identifier.
+    console.log('>>>>>>>', 'setting deserialize function');
+    passport.deserializeUser((id: string, done: Function) => {
+      deserialize(id, done);
+    });
+  };
+
+  public authenticate(strategy: string, opts: Object) {
```
tomjackman: Maybe this could be called `login` instead of `authenticate` to stay inline with keycloak naming scheme. 

We could have just a `login()` action which **just** performs authentication.

Then we can also have the `protect()` action (logic added down the line) which will do a few things:
1. Check if User is Authenticated, if not call `login()`
2. If user is authenticated, call `hasResourceRole(roleOnTheRoute)`
3. If user hadn't the role deny access, if they have the role, continue.

tomjackman (commented):
Good progress so far. Left a few comments.

demo/server/src/auth/local.ts
JameelB: it will be the same as `keycloak.protect()`. it will check if a user was authenticated and I can also add a check to check for roles as well. Will change the name to `protect` instead. I'm not sure what the `login-required` for keycloak does?

demo/server/src/app.ts
JameelB: No it's not necessary to have all fields in the req.user, you can choose which fields you'd like to be in it. I think the `getProfileData()` will just return what's in `req.user` unless we explicitly make it to read from the data source.

demo/server/src/auth/local.ts
tomjackman: Cool, just checking as the naming was ringing bells with keycloak.

login-required in keycloak just enforces a user to be logged into keycloak. ie. redirect to login if they are not authenticated, and leave them alone if authenticated, it just handles the auth mechanism side of things really.

demo/server/src/app.ts
tomjackman: I think a +1 to read from the data source rather than from the `req.user` to get the profile data. Best to have the data store be the source of truth, as we can return the user's roles and any other information using the same function. The req.user might get a bit bloated with other user metadata too.

demo/server/src/app.ts
```diff
@@ -1,33 +1,112 @@
...
 import * as path from 'path';
 import * as logger from 'morgan';
 import * as cookieParser from 'cookie-parser';
 import * as bodyParser from 'body-parser';
 import index from './routes/index';
 import * as favicon from 'serve-favicon';
 import EnvironmentConfig, { Config, CloudAppConfig } from './util/config';
+// import * as passport from 'passport';
+// import {Strategy} from 'passport-local';
+import users, {User} from './userSeed';
```
wtrocki: Get that wrapped into separate interface which can be implemented by users/use our store or in this particular case use list of seeded users.

demo/server/src/app.ts
```diff
@@ -1,33 +1,112 @@
...
 const app: express.Express = express();
 const appConfig: Config<CloudAppConfig> = new EnvironmentConfig<CloudAppConfig>();
 const config = appConfig.getConfig();
+// const authService = new PassportAuthImpl<User>();
 
 app.use(logger(config.morganOptions));
 app.use(bodyParser.json());
 app.use(bodyParser.urlencoded({ extended: false }));
 app.use(cookieParser());
+app.use(session({secret: 'test', resave: false, saveUninitialized: false}));
```
wtrocki: [Hint] if there are some strict requirements for top level app that aren't part of the module try to wrap them into separate class (similar to webapi aproach) - this way users will clearly know what elements they need to add and you can write documentation snippents pointing to this rather than maintaining chunks of data in docs.

demo/server/src/app.ts
```diff
@@ -1,33 +1,112 @@
...
 import EnvironmentConfig, { Config, CloudAppConfig } from './util/config';
+// import * as passport from 'passport';
+// import {Strategy} from 'passport-local';
+import users, {User} from './userSeed';
+import PassportAuthImpl from '@raincatcher/passport-auth';
 
 const app: express.Express = express();
 const appConfig: Config<CloudAppConfig> = new EnvironmentConfig<CloudAppConfig>();
 const config = appConfig.getConfig();
+// const authService = new PassportAuthImpl<User>();
```
wtrocki: I needed to uncomment that locally to compile.

demo/server/src/app.ts
```diff
@@ -1,33 +1,112 @@
...
 app.use('/', index);
 
 app.use((req: express.Request, res: express.Response, next) => {
-  var err: any = new Error('Not Found');
+  let err: any = new Error('Not Found');
   err.status = 404;
   next(err);
 });
 
+//passport config
```
wtrocki: use /* style comments - avoid making any changes in app.ts - use separate classes or pull out this to module as sample implementation.

demo/server/src/app.ts
wtrocki: you can use underscore to retrieve set of fields you want. 
For this particular example this code has no meaning. Let's keep it simple to improve readability of the examples.

demo/server/src/app.ts
```diff
@@ -1,33 +1,112 @@
...
+        avatar: user.avatar,
+        banner: user.banner,
+        notes: user.notes
+      };
+      done(null, userObj);
+    }
+  });
+});
+
+authService.setStrategy(function(username: string, password: string, done: Function) {
```
wtrocki: Can you justify using custom strategy? Just curios what lead you this way and what LocalStrategy was missing.

demo/server/src/app.ts
```diff
@@ -1,33 +1,112 @@
...
+  let user = users.map(function(user) {return user.username}).indexOf(username);
```
wtrocki: I do not like that we actually implying here some specific fields. Can we relay on actual methods on user facade interface `getLogin()` and `getPassword()`?

demo/server/src/app.ts
```diff
@@ -1,33 +1,112 @@
...
+
+  if(users[user] && users[user].password === password) {
```
wtrocki: delagate to `UserRepostiory` - to perform operation on user. Do not imply any implementation.

 `UserRepostiory` interface will need to define set of the metods to access users that are required from Security module point of view. Both repository and this code should be a part of this module.


ping @paolobueno - another useful example where we will use repository class

cloud/passportauth/src/auth/PassportSetup.ts
```diff
@@ -0,0 +1,73 @@
...
+    app.use(passport.initialize());
+    app.use(passport.session());
+
+    this.setStrategy();
+    this.setSerializeUser();
+    this.setDeserializeUser();
+  }
+
+  public setStrategy() {
+    passport.use(new Strategy((username, password, done) => {
```
tomjackman: Is there any difference between this and Passports Local strategy, or is the local strategy missing some functionality that we need?

cloud/passportauth/src/user/UserSec.ts
```diff
@@ -0,0 +1,42 @@
...
+class UserSecService<T extends BaseUser> implements UserSec {
+  protected userApi: UserApiService<T>;
+
+  constructor(protected readonly userApiService: UserApiService<T>) {
+    this.userApi = userApiService;
+  }
+
+  public getLogin(id: string) {
+    return this.userApi.read(id).then((user) => {
+      return user.username;
```
tomjackman: In the `/auth/PassportSetup.ts` file, there is `this.userSec.getLogin(username).then((user) => { ......` which calls the function here. A username gets passed in during auth, and then the username gets returned from the getLogin() method here? Should this be `userId` maybe?

tomjackman (commented):
Looking Very Good. Nice separation of the logic is making it much easier to read.

cloud/passportauth/src/auth/PassportSetup.ts
JameelB: This is the passport local strategy. Similar to this example here: https://github.com/passport/express-4.x-local-example/blob/master/server.js

cloud/passportauth/src/user/UserSec.ts
JameelB: ah yes, that should be userId

cloud/passportauth/src/auth/PassportSetup.ts
tomjackman: Ah, ok. I now see that the strategy is defined up the top of the file instead (different from current docs). But I think this is fine too, or better even.

cloud/passportauth/src/auth/PassportSetup.ts
```diff
@@ -0,0 +1,73 @@
...
+
+class PassportSetup<T extends BaseUser> {
+  protected userSec: UserSecService<T>;
+
+  constructor(protected readonly userSecService: UserSecService<T>) {
+    this.userSec = userSecService;
+  }
+
+  public init(app: express.Express) {
+    app.use(session({secret: 'test', resave: false, saveUninitialized: false}));
```
tomjackman: Something along the lines of this could be used for express-session, it can be looked at down the line though, just once we don't forget to add in some basic cookie security.

```
{
      secret: process.env.SESSION_SECRET || 'wfm',
      resave: false,
      saveUninitialized: false,
      cookie: {
        secure: process.env.NODE_ENV !== 'development',
        httpOnly: true,
        path: '/'
      }
    }
```

wtrocki: That will be defined by user.

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -0,0 +1,54 @@
+import * as express from 'express';
+import * as passport from 'passport';
+import BaseUser from '../user/BaseUser';
+import UserSecService from '../user/UserSec';
+
+interface Auth {
+  protect(req: express.Request, res: express.Response, next: () => void): void; // protect/authenticate?
```
wtrocki: Actual method may be:
`protect(role:string) => (req: express.Request, res: express.Response, next: () => void) => void;`

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -0,0 +1,54 @@
...
+  login(opts?: object): void;
```
wtrocki: Client interface method in cloud.

cloud/passportauth/src/auth/PassportSetup.ts
tomjackman: Yes it will, but for the demo scenario, we should have a good working example.

cloud/passportauth/integration/UserDataSource.ts
```diff
@@ -0,0 +1,26 @@
+import * as Promise from 'bluebird';
+import * as _ from 'lodash';
+import {User} from './UserSeedData';
+
+interface DataSource<T> {
+  read(id: string): object;
+}
+
+export class UserDataSource<T extends User> implements DataSource<T> {
```
wtrocki: I will go with DataRepository as we using this pattern across the Raincatcher.
"DataSource" - Represents unstructurized DataSource
"DataRepository" - Pattern to define set of the functions that will return expected results.

cloud/passportauth/integration/UserDataSource.ts
```diff
@@ -0,0 +1,26 @@
+import * as Promise from 'bluebird';
+import * as _ from 'lodash';
+import {User} from './UserSeedData';
+
+interface DataSource<T> {
+  read(id: string): object;
```
wtrocki:
```typescript
interface User{
   getUserName():string
  getPassword():string
}
interface UserRepository() {
 findUserById(id: string): User
}
```

cloud/passportauth/integration/UserDataSource.ts
```diff
@@ -0,0 +1,26 @@
+import * as Promise from 'bluebird';
+import * as _ from 'lodash';
+import {User} from './UserSeedData';
+
+interface DataSource<T> {
```
wtrocki: Let's drop generic here if we do not use it.

cloud/passportauth/integration/UserDataSource.ts
wtrocki: That's just suggestion

wtrocki: Actually best to return promises for this type of interface.

cloud/passportauth/integration/UserSeedData.ts
```diff
@@ -22,7 +23,8 @@ const userSeedData: User[] = [
     password: '123',
     phone: '2657258272',
     position: 'Senior Truck Driver',
-    username: 'trever'
+    username: 'trever',
+    roles: []
```
wtrocki: We know that we going to support 'User' and 'Admin" role so best to set them here.

cloud/passportauth/integration/index.ts
```diff
@@ -0,0 +1,33 @@
...
+
+app.use(bodyParser.json());
+app.use(cors());
+passportSetup.init(app); // initialize passport
+
+app.get('/testSecureEndpoint', authService.protect('test', '/login'), (req: express.Request, res: express.Response) => {
+  res.json({routeName: '/testSecureEndpoint', msg: 'secured route'});
+});
+
+app.get('/login', (req: express.Request, res: express.Response) => {
```
wtrocki: In examples this route normally returns page. Maybe we should add todo here?

cloud/passportauth/integration/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  res.json({routeName: '/login', msg: 'You need to login'});
+});
+
+app.post('/login', authService.login('local'), (req: express.Request, res: express.Response) => {
```
wtrocki: `authService.login('local')` => `authService.protect()`

I think it's fair to assume that users will use the same strategy for entire application. If we want to pass some additional information we can use following interface:

`authService.protect(role: string[], metadata?:any)`

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -1,54 +1,51 @@
...
 import * as passport from 'passport';
-import BaseUser from '../user/BaseUser';
-import UserSecService from '../user/UserSec';
+import {BaseUser} from '../user/BaseUser';
+import {UserSecService} from '../user/UserSec';
 
 interface Auth {
-  protect(req: express.Request, res: express.Response, next: () => void): void; // protect/authenticate?
-  login(opts?: object): void;
+  login(strategy: string, opts?: object): void;
```
wtrocki: [Question] What's the login here?

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -1,54 +1,51 @@
...
+  protect(role: string, loginRoute: string): void;
```
wtrocki: [Question] What's the purpose of `loginRoute`

cloud/passportauth/integration/UserSeedData.ts
```diff
@@ -0,0 +1,70 @@
+export interface User {
```
wtrocki: We can include here methods to map username and password in order to not hardcode them :D

cloud/passportauth/integration/UserSeedData.ts
```diff
@@ -0,0 +1,70 @@
...
+    banner: 'http://web18.streamhoster.com/pentonmedia/beefmagazine.com/TreverStockyards_38371.jpg',
+    email: 'trever@wfm.com',
+    id: 'rkX1fdSH',
+    name: 'Trever Smith',
+    notes: 'Trever doesn\'t work during the weekends.',
+    password: '123',
+    phone: '2657258272',
+    position: 'Senior Truck Driver',
+    username: 'trever',
+    roles: []
```
wtrocki: My previous comment is gone. I think we can add here actual roles for future.

cloud/passportauth/integration/index.ts
```diff
@@ -0,0 +1,33 @@
...
+const userSec = new UserSecService<User>(userApi);
+const passportSetup = new PassportSetup<User>(userSec);
+const authService = new PassportAuth<User>(userSec);
+const app = express();
+
+app.use(bodyParser.json());
+app.use(cors());
+passportSetup.init(app); // initialize passport
+
+app.get('/testSecureEndpoint', authService.protect('test', '/login'), (req: express.Request, res: express.Response) => {
```
wtrocki: I think it's fair to assume that users will use the same strategy for entire application. If we want to pass some additional information we can use following interface:

`authService.protect(role?: string[], metadata?:any)`

cloud/passportauth/integration/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  res.json({routeName: '/testSecureEndpoint', msg: 'secured route'});
+});
+
+app.get('/login', (req: express.Request, res: express.Response) => {
+  res.json({routeName: '/login', msg: 'You need to login'});
+});
+
+app.post('/login', authService.login('local'), (req: express.Request, res: express.Response) => {
+  res.json({routeName: '/login', msg: 'successfully logged in', user: req.user});
```
wtrocki: [Question] Do we need return json here or redirect?

cloud/passportauth/integration/index.ts
tomjackman: For 1.0, we will just use `authService.protect()` with just one param - a string of the role to protect the route with. 
eg. `authService.protect('testRole')`

cloud/passportauth/integration/index.ts
tomjackman: This should probably redirect to the HTTP referer.

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -0,0 +1,51 @@
...
+import {BaseUser} from '../user/BaseUser';
+import {UserSecService} from '../user/UserSec';
+
+interface Auth {
+  login(strategy: string, opts?: object): void;
+  protect(role: string, loginRoute: string): void;
+  isAuthenticated(req: express.Request): boolean;
+  getUserProfile(userId: string): void;
+  logout(req: express.Request): void;
+}
```
tomjackman: Can also add `hasResourceRole(role: string): void;`

cloud/passportauth/src/auth/PassportAuth.ts
JameelB: This is just passport.authenticate(). It doesn't handle any role checks or if user is already authenticated. It will just authenticate the user. The reason why it's called login here is because the equivalent of this is login() in keycloak @TommyJ1994

cloud/passportauth/src/auth/PassportAuth.ts
JameelB: This is a param for the login route to redirect to if the user wasn't logged in yet. I was thinking of having some kind of config for this instead of being passed as a param?

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  public getUserProfile(userId: string) {
+    return this.userSec.getProfileData(userId);
+  }
+
+  public logout(req: express.Request) {
+    req.logout();
+    // add a redirect here to homepage/logoutRoute/etc.?
+  }
+}
```
tomjackman:
```
public hasResourceRole(role: string) {
 // TODO Implementation Later
 }
```

cloud/passportauth/integration/index.ts
JameelB: Redirect would be best. The json there was just a temporary placeholder for me.

cloud/passportauth/src/auth/PassportAuth.ts
tomjackman: Yeah, we can remove the route param, as the role will be checked against the user, rather than the route.

cloud/passportauth/integration/index.ts
wtrocki: We may need to use what's needed to cover all requirements and options. In this particular example `/login` should be set globally and users should be always redirected to this global endpoint. I do not see any real use case where app will serve 2 different login endpoints.

wtrocki: @vchepeli  That is example of the functional test we build. We will need to have some tests that will check this endpoint

cloud/passportauth/integration/index.ts
wtrocki: Actually maybe we can just put there login page from example application. Let's use handlebars.
Having login page will allow us to perform end to end tests.

cloud/passportauth/src/auth/PassportAuth.ts
wtrocki: That's cloud interface. Some of the methods here are from client and will need to be removed later.

cloud/passportauth/src/auth/PassportAuth.ts
paolobueno: there's the `express.RequestHandler` available for these types too

cloud/passportauth/integration/index.ts
wtrocki: ping @spriadka

cloud/passportauth/integration/UserApi.ts
```diff
@@ -0,0 +1,40 @@
+import {BaseUser} from '../src/index';
+import {DataRepository} from './UserDataRepository';
+import {User} from './UserSeedData';
+
+/**
+ * A sample implementation of a base user api
+ */
+export class UserApi implements BaseUser {
```
wtrocki: As we agreed before. This entire class is not really needed as it serve no purpose. 
We can:
Have user interface with set of requirements as methods:
```
getUserById
```

Have single method in repository to get user object.

cloud/passportauth/integration/UserDataRepository.ts
```diff
@@ -0,0 +1,36 @@
...
+import {User} from './UserSeedData';
+
+export interface DataRepository<T> {
+  getUserById(id: string): Promise<T|undefined>;
+}
+
+/**
+ * A sample implementation of a user data repository
+ */
+export class UserDataRepository implements DataRepository<User> {
```
wtrocki: No need to have user as you imply that this will be user by name.

cloud/passportauth/integration/UserDataRepository.ts
```diff
@@ -0,0 +1,36 @@
+import * as Promise from 'bluebird';
+import * as _ from 'lodash';
+import {User} from './UserSeedData';
+
+export interface DataRepository<T> {
```
wtrocki: **Sorry for repeating the same comments but previous comment wasn't taken into account.**

Drop generic and add User as prefix to name. You building your own case for user. 
You have only one object that needs to be fetched from db. 
We can have RoleRepository - but I think that it's overcomplicating things. Just stick with user.

cloud/passportauth/integration/UserSeedData.ts
```diff
@@ -0,0 +1,70 @@
...
+  password: string;
+  phone: string;
+  position: string;
+  username: string;
+  roles: string[];
+}
+
+const userSeedData: User[] = [
+  {
+    avatar: 'https://s3.amazonaws.com/uifaces/faces/twitter/kolage/128.jpg',
```
wtrocki: As agreed let's remove this from module and provide small simple example. 
We going to integrate this dataset in mobile application.

cloud/passportauth/integration/public/login.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
```
wtrocki: 👍 ❤️

cloud/passportauth/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@raincatcher/passport-auth",
```
wtrocki: WDYT about`auth-passport`? This will help us to do some operations using lerna filter `auth-*`

cloud/passportauth/integration/UserApi.ts
wtrocki: As we agreed before. This entire class is not really needed as it serve no purpose. 
We can:
- Have BaseUser interface with set of requirements as methods:
```
user.getId
user.getPassword
user.getLogin
user.getRole
```

In this particular example:
getId implementation will be:

`return this.myCustomId`

Interface is implemented by *module integrators* - **mapping**

- Have single method in repository to get user object. 
```
getUserByLogin
```

I have no strong feelings to have this but you may find this really inflexible later when implementing actual solution. This interface may also involve some performance problems.

cloud/passportauth/integration/UserApi.ts
```diff
@@ -0,0 +1,40 @@
...
+  protected userDataRepo: DataRepository<User>;
+
+  constructor(protected readonly dataRepo: DataRepository<User>) {
+    this.userDataRepo = dataRepo;
+  }
+
+  public getId(loginId: string) {
+    return this.dataRepo.getUserById(loginId).then((user) => {
```
wtrocki: You can do `return this.getUsername();`

cloud/passportauth/integration/UserSeedData.ts
```diff
@@ -0,0 +1,70 @@
+export interface User {
+  avatar: string;
```
ghost: `avatar` should be optional  avatar?: string;

cloud/passportauth/integration/UserSeedData.ts
```diff
@@ -0,0 +1,70 @@
...
+  banner: string;
```
ghost: `banner` should be optional banner?: string;

cloud/passportauth/integration/UserSeedData.ts
```diff
@@ -0,0 +1,70 @@
...
+  email: string;
+  id: string;
+  name: string;
+  notes: string;
+  password: string;
+  phone: string;
+  position: string;
```
ghost: `position` should be optional position?: string;

cloud/passportauth/integration/UserSeedData.ts
ghost: this should be in examples folder. Core module should not contain any dummy data

cloud/passportauth/integration/index.ts
```diff
@@ -0,0 +1,53 @@
...
+import * as express from 'express';
+import * as path from 'path';
+import passport, {PassportAuth, PassportSetup, UserSecService} from '../src/index';
+import UserApi from './UserApi';
+import UserDataRepository from './UserDataRepository';
+import userSeedData from './UserSeedData';
+
+// Configuration for express session options
+const sessionOpts = {
+  secret: process.env.SESSION_SECRET || 'wfm',
```
ghost: should we still use wfm keywords?

cloud/passportauth/integration/index.ts
```diff
@@ -0,0 +1,53 @@
...
+const app = express();
+
+app.use(bodyParser.json());
+app.use(bodyParser.urlencoded({
+  extended: false
+}));
+app.use(cors());
+passportSetup.init(app, sessionOpts);
+
+app.get('/testAdminEndpoint', authService.protect('admin'), (req: express.Request, res: express.Response) => {
```
ghost: why its `/test*Endpoint`? For testing purposes?

cloud/passportauth/src/auth/PassportSetup.ts
```diff
@@ -0,0 +1,79 @@
...
+   * Sets the local strategy used by passport authentication.
+   */
+  public setStrategy() {
+    passport.use(new Strategy((loginId, password, done) => {
+      this.userSec.getUserId(loginId).then((userId: string) => {
+        if (!userId) {
+          return done(null, false);
+        } else {
+          this.userSec.comparePassword(userId, password).then((valid: boolean) => {
+            return valid ? done(null, userId) : done(null, false);
```
ghost: Here is some type inconsistency `done(null, userId) : done(null, false);` one is userId: string and other is false: boolean

cloud/passportauth/src/auth/PassportSetup.ts
```diff
@@ -0,0 +1,79 @@
...
+          });
+        }
+      })
+      .catch((err: Error) => {
+        return done(err);
```
ghost: better to log this error with log.error()

cloud/passportauth/README.md
```diff
@@ -0,0 +1,75 @@
+# RainCatcher Passport Auth
+Raincatcher authentication and authorization module using [PassportJS][1]. This module creates and initializes a Passport authentication service using Passport's local strategy.
+This allows express routes to be protected from requests by checking if the user is authenticated and/or authorized with the required role. This module also
```
wtrocki: This allows => Component protects

wtrocki: "This module also"

If there is more than 3 responsibilities (and I think there are) list list them using bullet points
Use common facts and avoid long sentences.

cloud/passportauth/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+2. **Initialize Passport auth service**
+   ```typescript
+       ...
+       const userSec = new UserSecService(userApi);
+       const passportSetup = new PassportSetup(userSec);
+       const authService = new PassportAuth(userSec);
+       ...
+   ```
+    - UserSecService
```
wtrocki: Having such long description here may be hard to maintain. This stuff should be part of the TypeDoc we going to generate. Let's move this to documentation in code - that's easier to maintain. Here we can just say what class is doing and refer users to documentation/code assist etc.

Reason why we do this is just simple maintenance - I guarantee you this will be quickly outdated and no longer reference actual implementation

cloud/passportauth/README.md
```diff
@@ -0,0 +1,75 @@
...
+        - Params:
+            - UserApi **[Required]**: Implementation of the BaseUser interface that is used to retrieve user data (id, login id, password hash and roles)
```
wtrocki: Name is really missleading.  UserRepository sounds better here.

cloud/passportauth/README.md
```diff
@@ -0,0 +1,75 @@
...
+    - PassportAuth
+        - Params:
+            - UserSec **[Required]**
+            - LoginRoute **[Optional]**: The login route used to redirect to if the user is not authenticated. Default value: `'/login'`
+
+3. **Setup Passport**
+    ```typescript
+        ...
+        // Configuration for express session options
+        const sessionOpts = {
```
wtrocki: let's not put actual session options but reference users to documentation.

cloud/passportauth/README.md
```diff
@@ -0,0 +1,75 @@
...
+            path: '/'
+          }
+        };
+        ...
+
+        passportSetup.init(app, sessionOpts);
+    ```
+    - passportSetup.init(): This will initialize express session and passport, as well as set the strategy, serializeUser and deserializeUser
+    functions to be used by passport.
+        - params:
```
wtrocki: Same as above.

cloud/passportauth/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+    See [express-session][2] for more information on the available express session options.
+
+#### Usage
+```typescript
+    app.get('/secureEndpoint', authService.protect('admin'), (req: express.Request, res: express.Response) => {
+      res.json({routeName: '/secureEndpoint', msg: 'authenticated and authorized to access secure resource'});
+    });
+```
+  - authService.protect()
```
wtrocki: Same as above.

cloud/passportauth/example/public/login.html
```diff
@@ -0,0 +1,73 @@
+<html xmlns="http://www.w3.org/1999/xhtml" class="">
+
+<head>
+  <meta charset="utf-8">
+  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+  <meta name="robots" content="noindex, nofollow">
+
+  <title>        Log in to WFM
```
ghost: WFM is used?

cloud/passportauth/example/public/login.html
```diff
@@ -0,0 +1,73 @@
...
+</head>
+
+<body class="">
+<div id="kc-logo"><a href="#"><div id="kc-logo-wrapper"></div></a></div>
+
+<div id="kc-container" class="">
+  <div id="kc-container-wrapper" class="">
+
+    <div id="kc-header" class="">
+      <div id="kc-header-wrapper" class="">        Feedhenry WFM
```
ghost: same as above

cloud/passportauth/example/UserApi.ts
```diff
@@ -0,0 +1,37 @@
+import * as Promise from 'bluebird';
+import {BaseUser} from '../src/index';
+import UserDataRepository from './UserDataRepository';
+import {User} from './UserSeedData';
+
+/**
+ * A sample implementation of a base user api
+ */
+export class UserApi implements BaseUser {
```
wtrocki: Let's correct this as we agreed. See outdated comment.

cloud/passportauth/example/UserDataRepository.ts
```diff
@@ -0,0 +1,36 @@
...
+    this.userData = seedData;
+  }
+
+  /**
+   * A sample get user using an id from a data source
+   *
+   * @param id {string} - A unique id used to identify the user
+   * @returns {Promise} - Returns a user object if user was found
+   */
+  public getUserById(id: string) {
```
wtrocki: I think we looking for login name. Implementors can actually decide what it will be.

cloud/passportauth/example/public/login.html
wtrocki: Good point. We should template this as {PRODUCT NAME} and pass it from config.

cloud/passportauth/example/public/login.html
```diff
@@ -0,0 +1,73 @@
...
+  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+  <meta name="robots" content="noindex, nofollow">
+
+  <title>        Log in to WFM
+  </title>
+  <link href="style.css" rel="stylesheet" />
+</head>
+
+<body class="">
+<div id="kc-logo"><a href="#"><div id="kc-logo-wrapper"></div></a></div>
```
wtrocki: Replace `kc` with login or something that will be less keycloak related :)

cloud/passportauth/example/public/login.html
```diff
@@ -0,0 +1,73 @@
...
+    </div>
+
+
+    <div id="kc-content" class="">
+      <div id="kc-content-wrapper" class="">
+
+
+        <div id="kc-form" class="">
+          <div id="kc-form-wrapper" class="">
+            <!-- TODO some login route -->
```
wtrocki: Remove TODO

cloud/passportauth/example/public/login.html
```diff
@@ -0,0 +1,73 @@
...
+
+              <div class="">
+                <div id="kc-form-options" class="">
+                  <div class="">
+                  </div>
+                </div>
+
+                <div id="kc-form-buttons" class="">
+                  <div class="">
+                    <input class="  " name="login" id="kc-login" type="submit" value="Submit"/>
```
wtrocki: remove empty class objects.

cloud/passportauth/example/public/style.css
```diff
@@ -0,0 +1,198 @@
+/* @import url('../../keycloak/lib/zocial/zocial.css');*/
+
+body {
+  background-color: rgb(63,81,181);
+  /*background-image: url('../img/bkgrnd.jpg');*/
```
wtrocki: Remove all comments.

cloud/passportauth/example/public/style.css
```diff
@@ -0,0 +1,198 @@
...
+  margin: 0px;
+}
+
+a {
+  color: #fff;
+}
+
+/* Change content to change the title of the page*/
+div#kc-header::after {
+  content: 'FeedHenry Work Force Management';
```
wtrocki: Avoid this in CSS

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -0,0 +1,45 @@
...
+   * @param role {string} [Optional] Role which the user needs in order to access this resource
+   */
+  public protect(role?: string) {
+    return (req: express.Request, res: express.Response, next: express.NextFunction) => {
+      if (req.isAuthenticated()) {
+        if (!role) {
+          return next();
+        } else {
+          this.userSec.hasResourceRole(role).then((hasRole: boolean) => {
+            return hasRole ? next() : res.status(401).send('Unauthorized');
```
wtrocki: [Antipattern] Avoid sending strings like `'Unauthorized'` etc. This may break json parsers. 
Returning 401 it's enough. Body can be empty.

cloud/passportauth/src/auth/PassportSetup.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    this.setStrategy();
+    this.setSerializeUser();
+    this.setDeserializeUser();
+  }
+
+  /**
+   * Sets the local strategy used by passport authentication.
+   */
+  public setStrategy() {
```
wtrocki: As agreed - let's allow people to setup this. 

Expose our strategy as constant so people can do something like:

`setStrategy(DEFAULT_STRATEGY)`

cloud/passportauth/src/auth/PassportSetup.ts
```diff
@@ -0,0 +1,79 @@
...
+  public setSerializeUser() {
+    passport.serializeUser((userId: string, done) => {
+      return done(null, userId);
+    });
+  }
+
+  /**
+   * Sets the deserializeUser function to be used by passport.
+   */
+  public setDeserializeUser() {
```
wtrocki: As we spoke: We should not make decisions here what is being serialized/deserialized. That should be part of the API.

cloud/passportauth/src/user/UserSec.ts
```diff
@@ -0,0 +1,52 @@
+import BaseUser from './BaseUser';
+
+export interface UserSec {
+  getUserId(loginId: string): Promise<string|null>;
+  comparePassword(password: string): Promise<boolean>;
+  hasResourceRole(role: string): Promise<boolean>;
+}
+
+export class UserSecService implements UserSec {
+  protected userApi: BaseUser;
```
wtrocki: As above - let's not use `userApi` - `user` will be cleaner here.

cloud/passportauth/src/user/UserSec.ts
```diff
@@ -0,0 +1,52 @@
...
+    this.userApi = UserApi;
+  }
+
+  /**
+   * Retrieves the user's id using the user's login id.
+   *
+   * @param loginId {string} - A unique login id
+   * @returns {Promise<string>} - Returns the user's id if found
+   */
+  public getUserId(loginId: string) {
```
wtrocki: Why this is needed? IMHO bringing id to this it's a problem. 
We should relay on 3 fields: username, password and role. Username is unique identifier itself. Id is really making our implementation overly complicated.

cloud/passportauth/example/public/login.html
```diff
@@ -0,0 +1,73 @@
...
+                </div>
+              </div>
+
+              <div class="">
+                <div class="">
+                  <label for="password" class="">Password</label>
+                </div>
+
+                <div class="">
+                  <input id="password" class="" name="password" type="password" autocomplete="off" />
```
tomjackman: `autocomplete="user-password"` - TL;DR - Some browsers/versions ignore the `off` value.

demo/server/src/app.ts
```diff
@@ -20,7 +19,7 @@ app.use(cookieParser());
 app.use(express.static(path.join(__dirname, 'public')));
 app.use(favicon(path.join(__dirname, '../public', 'favicon.ico')));
 
-app.use('/', index);
+// app.use('/', index);
```
wtrocki: Remove this

cloud/passportauth/README.md
```diff
@@ -0,0 +1,75 @@
+# RainCatcher Passport Auth
+Raincatcher authentication and authorization module using [PassportJS][1]. This module creates and initializes a Passport authentication service using Passport's local strategy.
```
paolobueno: We can just inline links instead of using number references, a bit easier for maintenance. Could add a link to [local strategy](https://github.com/jaredhanson/passport-local) and decide whether to call the lib Passport or PassportJS.

cloud/passportauth/README.md
paolobueno: 'authenticated and/or authorized' needs better distinction, they're not synonims.

authenticated is just "I know you you are", authorized is "based on who you are, you are (not) allowed to access this"

cloud/passportauth/README.md
```diff
@@ -0,0 +1,75 @@
...
+## Quick Start
+#### Setup & Pre-requisites
+1. **Implement BaseUser Interface**
+
+    Implement the BaseUser interface defined in `./src/user/BaseUser.ts` in order to allow for retrieving the required user data from your data source.
+    See `./example/UserApi.ts` for a sample implementation of this interface
+
+2. **Initialize Passport auth service**
+   ```typescript
+       ...
```
paolobueno: use the triple bactick but not the four spaces, this makes the code indented by 4 spaces.

cloud/passportauth/README.md
```diff
@@ -0,0 +1,75 @@
...
+Raincatcher authentication and authorization module using [PassportJS][1]. This module creates and initializes a Passport authentication service using Passport's local strategy.
+This allows express routes to be protected from requests by checking if the user is authenticated and/or authorized with the required role. This module also
+uses persistent login sessions using [express-session][2].
+
+
+## Quick Start
+#### Setup & Pre-requisites
+1. **Implement BaseUser Interface**
+
+    Implement the BaseUser interface defined in `./src/user/BaseUser.ts` in order to allow for retrieving the required user data from your data source.
```
paolobueno: class references can be links: [`BaseUser`](src/user/BaseUser.ts)

(edit this comment to see markdown source)

cloud/passportauth/README.md
```diff
@@ -0,0 +1,75 @@
...
+        - Params:
+            - UserSec **[Required]**
+    - PassportAuth
+        - Params:
+            - UserSec **[Required]**
+            - LoginRoute **[Optional]**: The login route used to redirect to if the user is not authenticated. Default value: `'/login'`
+
+3. **Setup Passport**
+    ```typescript
+        ...
```
paolobueno: another case for 'no four spaces'

paolobueno: actually this is using 4+4 spaces? :joy:

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -0,0 +1,45 @@
...
+    this.userSec = UserSecService;
+    this.loginRoute = loginRoute || '/login';
+  }
+
+  /**
+   * Function which checks if the user requesting access to the resource is authenticated and authorized to
+   * access the resource. Redirects to the login page if user is not authenticated or returns an 'Unauthorized'
+   * response if user does not have the required role.
+   *
+   * @param role {string} [Optional] Role which the user needs in order to access this resource
```
paolobueno: no need to add `{string} [Optional]`, those comes from the annotations

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -0,0 +1,45 @@
...
+   */
+  public protect(role?: string) {
+    return (req: express.Request, res: express.Response, next: express.NextFunction) => {
+      if (req.isAuthenticated()) {
+        if (!role) {
+          return next();
```
paolobueno: could add a `console.warn` here when skips happen, very hard to track otherwise

it also serves as a todo for when we include hte logger module

cloud/passportauth/src/user/BaseUser.ts
```diff
@@ -0,0 +1,32 @@
+export interface BaseUser {
```
paolobueno: -1 for naming interfaces `Base*`, these should be base implementations designed to be extended and provide part of the functionality the interface requires

cloud/passportauth/src/auth/PassportAuth.ts
wtrocki: Logger module is actually available.

cloud/passportauth/src/user/BaseUser.ts
paolobueno: See the current [BaseTask](https://github.com/feedhenry-raincatcher/raincatcher-core/blob/vehicle-inspection-example/cloud/wfm/src/task/BaseTask.ts) for an example, it implements the `Task` interface just with the intent of making it easier for users to override the parts they need to.

cloud/passportauth/src/auth/PassportAuth.ts
paolobueno: yeah, PR can be updated to use it or left to a follow up issue, no strong opinion from me

cloud/passportauth/example/UserApi.ts
```diff
@@ -0,0 +1,37 @@
+import * as Promise from 'bluebird';
+import {BaseUser} from '../src/index';
+import UserDataRepository from './UserDataRepository';
+import {User} from './UserSeedData';
+
+/**
+ * A sample implementation of a base user api
```
paolobueno: Is this really just an example? We plan on supplying supported implementations of interfaces, so maybe this would be it?

cloud/passportauth/example/UserApi.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export class UserApi implements BaseUser {
+  protected userDataRepo: UserDataRepository;
+  protected user: User|undefined;
+
+  constructor(protected readonly dataRepo: UserDataRepository) {
+    this.userDataRepo = dataRepo;
```
paolobueno: This is redundant, when you add `protected|public` to a constructor param it already creates a member in the class, so here you have `this.userDataRepo === this.dataRepo`, can remove one of them.

tomjackman:
@JameelB Just a heads up. I think it is important that this PR doesn't try and 'achieve' too much, it looks like it might be getting a bit out of control. There are 9 other tickets in this epic still in Open state, but judging by the checklist in the PR description, it looks like some of these have been partially covered already in this PR. I think its Important to keep the PR related to the ticket that is being worked on, otherwise gauging progress/work effort remaining is difficult and finer details in the Open tickets can be overlooked.

cloud/passportauth/example/public/login.html
ghost: Replace kc- with rc- like 'raincatcher'

cloud/passportauth/src/auth/PassportAuth.ts
ghost: res.status(401).send(new Error('Unauthorized')); to be more specific

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -0,0 +1,45 @@
...
+        } else {
```
paolobueno: no need to have `else` in short-circuits

cloud/passportauth/src/auth/PassportSetup.ts
```diff
@@ -0,0 +1,79 @@
...
+   * Initializes passport authentication on an Express application.
+   * This initializes the Express session, passport as well as
+   * setting the local strategy, serializeUser and deserializeUser function.
+   *
+   * @param app {express.Express} An express app used by the application.
+   * @param sessionOpts {SessionOptions} Configuration used by Express session
+   */
+  public init(app: express.Express, sessionOpts: SessionOptions) {
+    app.use(session(sessionOpts));
+    app.use(passport.initialize());
```
ghost: should not be called without app.use()? just `passport.initialize()`

cloud/passportauth/src/auth/PassportAuth.ts
paolobueno: Remove brackets from single-statement arrow functions.

btw do you need the `: boolean` annotation here? hopefully it should come from the generic `.then<T>`

cloud/passportauth/src/auth/PassportAuth.ts
paolobueno: in general try to simplify the branching in this function

cloud/passportauth/src/user/BaseUser.ts
ghost: @JameelB UserAPI should be interface and BaseUser should be base implementation on UserAPI interface as @paolobueno  mentioned above

cloud/passportauth/src/auth/PassportSetup.ts
tomjackman: @vchepeli No, its to be used as middleware.

cloud/passportauth/src/user/BaseUser.ts
```diff
@@ -0,0 +1,32 @@
...
+   * @returns {Promise<string>} - A unique id used to identify the user
+   */
+  getId(loginId: string): Promise<string|null>;
+
+  /**
+   * Retrieves the user's login id
+   *
+   * @returns {Promise<string>} - A unique id used by the user for login
+   */
+  getLoginId(): Promise<string|null>;
```
vchepeli: should be stateless; you need to pass any user to interface
getLoginId(user) {return user.id}

cloud/passportauth/src/user/BaseUser.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * Retrieves the user's hashed password
+   *
+   * @returns {Promise<string>} - User's hashed password
+   */
+  getPasswordHash(): Promise<string|null>;
```
vchepeli: should be stateless; you need to pass any user to interface
getPasswordHash(user) {return user.password}

cloud/passportauth/src/user/BaseUser.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * Retrieves the user's roles
+   *
+   * @returns {Promise<string[]>} - An array containing roles assigned to the user
+   */
+  getRoles(): Promise<string[]|null>;
```
vchepeli: should be stateless; you need to pass any user to interface
getRoles(user) {return user.roles}

cloud/passportauth/src/auth/PassportAuth.ts
tomjackman: It could be done as part of the `hasResourceRole()` function in `cloud/passportauth/src/user/UserSec.ts` too.

cloud/passportauth/src/auth/PassportSetup.ts
```diff
@@ -0,0 +1,79 @@
...
+import {SessionOptions} from 'express-session';
+import * as passport from 'passport';
+import {Strategy} from 'passport-local';
+import {UserSec} from '../user/UserSec';
+
+export interface Setup {
+  init(app: express.Express, sessionOpts: SessionOptions): void;
+  setStrategy(): void;
+  setSerializeUser(): void;
+  setDeserializeUser(): void;
```
paolobueno: These empty signatures are implementation details, can't enforce them through an interface.

These look like they would be better fitting as a set of options:
```typescript
serialize: (user: User, cb: (error: Error, id: string) => any) => any;
deserialize: (id: string, cb: (error: Error, user: User) => any) => any;
```

cloud/passportauth/test/PassportSetupTest.ts
```diff
@@ -0,0 +1,92 @@
...
+    mockRes = {
+      status: sinon.stub().returns({
+        send: sinon.spy()
+      }),
+      json: sinon.spy(),
+      redirect: sinon.spy(),
+      end: sinon.spy()
+    };
+    mockNext = sinon.spy();
+    done();
```
paolobueno: [optional] no need to use the done callback when your function is syncronous

cloud/passportauth/src/auth/PassportAuth.ts
wtrocki: Logger module is already available and really easy to setup.

cloud/passportauth/example/UserSeedData.ts
```diff
@@ -0,0 +1,23 @@
+export interface User {
```
wtrocki: That interface is not needed now.

cloud/passportauth/example/public/login.html
```diff
@@ -0,0 +1,73 @@
+<html xmlns="http://www.w3.org/1999/xhtml" class="">
```
wtrocki: As agreed let's use simple login in example and have proper shinny one in demo app.
Comments from here can be applied there.

ping @austincunningham

cloud/passportauth/example/UserDataRepository.ts
```diff
@@ -0,0 +1,36 @@
+import * as Promise from 'bluebird';
+import * as _ from 'lodash';
+import {User} from './UserSeedData';
+
+export interface UserDataRepo {
```
wtrocki: Let's move this to module code.

cloud/passportauth/example/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  saveUninitialized: false,
+  cookie: {
+    secure: process.env.NODE_ENV === 'production',
+    httpOnly: true,
+    path: '/'
+  }
+};
+
+// Initialize user data repository and passport
+const userRepo = new UserDataRepository(userSeedData);
```
wtrocki: Once we simplify this maybe we can have just all code in single file? WDYT?

cloud/passportauth/example/index.ts
```diff
@@ -0,0 +1,55 @@
...
+const userApi = new UserApi(userRepo);
+const userSec = new UserSecService(userApi);
+const passportSetup = new PassportSetup(userSec);
+const authService = new PassportAuth(userSec);
```
wtrocki: Only authService is being used in code bellow. Maybe we can try to condense this into single class.

wtrocki:
@TommyJ1994 - We covering here 3 different tickets (as work was also done by Austin).
We are close to finish this. Just minor comments were added to avoid breaking our interfaces once we merge that.

cloud/passportauth/src/auth/PassportAuth.ts
ghost: `return next()` should be last statement in function

wtrocki (dismissed):

wtrocki (approved):

PR #10, opened 2017-06-13
RAINCATCH-915 @raincatcher/wfm module interface definitions

paolobueno:
Feature branch for [RAINCATCH-915](https://issues.jboss.org/browse/RAINCATCH-915), can be PR'd against from branches for the other subtasks.

Renamed entities from existing RainCatcher codebase:

- Workflow -> Process
- Workorder -> ProcessInstance (Not part of BPMN, inspired by [Camunda](https://docs.camunda.org/manual/7.7/user-guide/process-engine/process-engine-concepts/), suggestions welcome)
- Step -> Task, new entity, renamed from spike

Result has the same name and task

- [x] Import definitions from initial spike
- [x] Rename to match BPMN entities
- [x] Import unit tests from spike
- [x] Modify tests' structure to match example-base (index.ts with reusable suite, implementation-specific tests in separate file)
- [x] Organize module's exports via `index.ts`
- [x] Define base interface for Process
- [x] Define base interface for ProcessInstance
- [x] Define base interface for Task
- [x] Define base interface for the Executor
- [x] Define base interface for Result
- [x] Define sample interface for custom Results
- [x] Add JsDocs
- [x] Refactor out default exports
- [x] Add complete usage example to Readme
- [x] Rename integration/ to example/
- [ ] Use logger module
- [x] Add possibility to fill `Process.tasks` via constructor


paolobueno:
ping @wtrocki @vchepeli @nialldonnellyfh @JameelB

paolobueno:
This is also missing a lot of the repository dependencies required for robust persistent storage.

My idea is generally having the interfaces expose async methods (returning Promises) and have the specific implementations require the needed repositories on their constructores.

cloud/wfm/src/process-instance/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  getId(): string;
+
+  /**
+   * Event emitted when
+   */
+  on(event: 'task:change', handler: (e: InstanceStepEventData<this>) => any): this;
+  on(event: 'process:done', handler: (e: InstanceEventData<this>) => any): this;
+}
+
+class InstanceImpl extends EventEmitter implements ProcessInstance {
```
wtrocki: EventEmitter ? Can you use implementation from raincatcher-typescript here.

wtrocki:
@paolobueno - Proposed work is really similar to things we have at the moment. 
I'm trying to adapt our existing *client side* code to typescript and I will use some pieces from here to do it - then we can compare this and see what are the differences. etc.

I'm particularly interested in execution and mapping views so actual UI will not be affected.

cloud/wfm/src/executor/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  process: Process;
+  instance?: ProcessInstance;
+  start(): void;
+}
+
+export interface InstanceRepository {
+  save(instance: ProcessInstance): Promise<ProcessInstance>;
+}
+
+class ExecutorImpl implements Executor {
```
wtrocki: 👍

cloud/wfm/src/process-instance/index.ts
paolobueno: It's this guy: https://github.com/primus/eventemitter3, runs everywhere and is compatible with nodejs' api

examples/base/test/mocha.opts
```diff
@@ -1,3 +1,3 @@
 --compilers ts:ts-node/register
 --require source-map-support/register
-test/**.ts
+test/**/*.ts
```
wtrocki: 👍

cloud/wfm/src/executor/index.ts
```diff
@@ -0,0 +1,45 @@
...
+  constructor(public process: Process, public instanceRepository: InstanceRepository) {
+    this.instance = this.process.createInstance();
+  }
+  public start() {
+    this.runCurrentTask();
+  }
+  protected onTaskDone(e: TaskEventData<Task>) {
+    if (e.currentStatus === TaskStatus.done) {
+      Promise.resolve(this.instance.next())
+        .then(() => this.saveInstance)
```
ghost: `.then(this.saveInstance)` ?

cloud/wfm/src/executor/index.ts
```diff
@@ -0,0 +1,45 @@
...
+        .then(() => this.runCurrentTask());
```
ghost: `.then(this.runCurrentTask)` ?

cloud/wfm/src/executor/index.ts
```diff
@@ -0,0 +1,45 @@
...
+  public instance: ProcessInstance;
+  constructor(public process: Process, public instanceRepository: InstanceRepository) {
+    this.instance = this.process.createInstance();
+  }
+  public start() {
+    this.runCurrentTask();
+  }
+  protected onTaskDone(e: TaskEventData<Task>) {
+    if (e.currentStatus === TaskStatus.done) {
+      Promise.resolve(this.instance.next())
```
ghost: `Promise.resolve(this.instance.next)` ?

cloud/wfm/src/process-instance/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  public id: string;
+  public assigneeId: string;
+  public processId: string;
+  public currentTask: Task;
+  public tasks: Task[];
+
+  protected currentStepIdx: number;
```
ghost: currentTaskIdx ?

cloud/wfm/src/process-instance/index.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  // TODO: add repository for data storage
+  constructor(initialSteps: Task[]) {
```
ghost: initialTasks ?

cloud/wfm/src/task/BaseTask.ts
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Base implementation for
+ */
+class BaseTask extends EventEmitter implements Task {
+  protected _status: TaskStatus | number = TaskStatus.pending;
+  protected options: object;
+
+  public set status(to: TaskStatus | number) {
+    const e: TaskEventData<this> = {
+      date: new Date(),
```
ghost: ` date: new Date(),` you want to track all status changes here? How often it can be tracked, otherwise it can lead to memory leaks

cloud/wfm/test/process-instance/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    it('next() should fire events related to a change in the active Task', function(done) {
+      instance.on('task:change', function(e) {
+        assert(e.instance === instance);
+        done();
+      });
+      instance.next();
+    });
+
+    it('should fire an event when all Tasks are done');
```
ghost: add //TODO comment to track all unimplemented tests

cloud/wfm/test/task/BaseTask.ts
```diff
@@ -0,0 +1,15 @@
...
+describe('BaseTask', function() {
+  let task: BaseTask;
+  beforeEach(function() {
+    task = new BaseTask();
+  });
+
+  suite(() => new BaseTask());
+
+  describe('run', function() {
+    it('should set the status to done');
```
ghost: add //TODO comment to track all unimplemented tests

cloud/wfm/test/task/index.ts
```diff
@@ -0,0 +1,18 @@
+import { Task } from '../../src/task';
+function suite(taskFactory: () => Task) {
+  describe('Task Interface', function() {
+    it('should emit a statusChange event on a changed status');
```
ghost: add //TODO comment to track all unimplemented tests

cloud/wfm/test/task/index.ts
```diff
@@ -0,0 +1,18 @@
...
+    describe('#getStatus', function() {
+      it('should round a custom status number to a value on the TaskStatus enum');
```
ghost: add //TODO comment to track all unimplemented tests

vchepeli (commented):
Unit tests missing. Other small changes required

cloud/wfm/src/task/ParallelTask.ts
```diff
@@ -0,0 +1,18 @@
...
+class ParallelTask extends BaseTask implements Task {
+  public status;
+  constructor(public tasks: Task[]) {
+    super();
+  }
+  public run() {
+    this.tasks.forEach(s => this.setupListening(s));
+  }
+
+  protected setupListening(step: Task) {
```
ghost: `task: Task` ?

cloud/wfm/src/executor/index.ts
paolobueno: the problem with passing the function is that it's not bound to `this`, so the equivalent is `.then(this.saveInstance.bind(this))`, though since we have arrow functions I think it's more readable with them.

cloud/wfm/src/process-instance/index.ts
paolobueno: :+1: nice catch, rename fail :joy:

cloud/wfm/src/task/BaseTask.ts
paolobueno: This is actually just reporting the time itself, so the tracking would be optional and left for client code.

The tracker would listen on these events and register them however it wants to, most likely to external persistent storage, anyway not something we should tackle now.

TL;DR we should design the apis so they can be properly observed by whoever wants to, but not care about why and how for the moment.

cloud/wfm/src/executor/index.ts
```diff
@@ -0,0 +1,45 @@
...
+  instance?: ProcessInstance;
+  start(): void;
+}
+
+export interface InstanceRepository {
+  save(instance: ProcessInstance): Promise<ProcessInstance>;
+}
+
+class ExecutorImpl implements Executor {
+  public instance: ProcessInstance;
```
ghost: think this one should have some other name; `processInstance = new Process()`, we can call it ProcessOrder or ProcessAck, WDYT @nialldonnellyfh, @wtrocki?

cloud/wfm/src/task/index.ts
```diff
@@ -0,0 +1,83 @@
...
+ *  The intervals between each of the values can be used by custom implementations
+ * for additional statuses
+ *
+ * Another advantage is that apart from 'blocked' in general steps are expected to move to higher-numbered statuses
+ */
+export enum TaskStatus {
+  /**
+   * Indicates that the {@link Step} has been created and is unnassigned
+   */
+  'pending' = 0,
```
ghost: think better use UPPER_CASE here like constants; PENDING, ASSIGNED, DONE...

cloud/wfm/test/process-instance/index.ts
paolobueno: When there's no function on the `it` it shows as pending on mocha's output. IMO that's enough of a TODO :slightly_smiling_face:

wtrocki:
[Style] I see you using `index.ts` for every folder as pattern while other modules always reference class/interface name.

cloud/wfm/src/executor/ExecutorImpl.ts
```diff
@@ -0,0 +1,33 @@
+import * as Promise from 'bluebird';
+import {Process} from '../process';
+import {ProcessInstance} from '../process-instance';
+import {Task, TaskEventData, TaskStatus} from '../task';
+import {Executor} from './index';
+
+export interface InstanceRepository {
```
wtrocki: Typical pattern is to have repositories in separate folder or at least separate file.

cloud/wfm/src/task/index.ts
```diff
@@ -0,0 +1,91 @@
...
+ *  The intervals between each of the values can be used by custom implementations
+ * for additional statuses
+ *
+ * Another advantage is that apart from 'blocked' in general steps are expected to move to higher-numbered statuses
+ */
+export enum TaskStatus {
+  /**
+   * Indicates that the Task has been created and is unnassigned
+   */
+  'pending' = 0,
```
wtrocki: Can we move this to upper case ?

cloud/wfm/src/result/UrlResult.ts
```diff
@@ -0,0 +1,18 @@
+// NOTE: browserify can provide an implementation for node's url module
+import * as nodeUrl from 'url';
+import {Result} from './index';
+
+/**
+ * Represents a {@link Result} that can be accessed via a URL
+ */
+export default class UrlResult implements Result {
```
wtrocki: IMHO this was good for experiments but has zero value for solution. Best to remove.

paolobueno:
Rebased on master

paolobueno:
Build failing due to lack of coverage. Implemented unit tests are green, lots of pending ones to be dealt with in follow up issues [924](https://issues.jboss.org/browse/RAINCATCH-924), [925](https://issues.jboss.org/browse/RAINCATCH-925), [926](https://issues.jboss.org/browse/RAINCATCH-926)

cloud/wfm/src/JsonSchema.ts
```diff
@@ -0,0 +1,155 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+// retrieved from https://gist.github.com/enriched/c84a2a99f886654149908091a3183e15
+
+interface JsonSchema {
```
wtrocki: I saw this in many PR's before. Is there any particular reason why we need that as source?
There are numerous npm modules like:
https://www.npmjs.com/package/jsonschema etc. that offers this.

cloud/wfm/package.json
```diff
@@ -0,0 +1,51 @@
...
+    ],
+    "require": [
+      "ts-node/register",
+      "source-map-support/register"
+    ],
+    "produce-source-map": true,
+    "check-coverage": true,
+    "lines": 75,
+    "functions": 100,
+    "branches": 80
```
wtrocki: Change limits to fix build ;)

cloud/wfm/src/executor/ExecutorImpl.ts
```diff
@@ -0,0 +1,37 @@
+import * as Promise from 'bluebird';
+import {Process} from '../process';
+import {ProcessInstance} from '../process-instance';
+import {Task, TaskEventData, TaskStatus} from '../task';
+import {Executor} from './index';
+import {InstanceRepository} from './InstanceRepository';
+
+export {InstanceRepository};
```
wtrocki: IMHO that's antipattern. File should export all elements that are inside.
Then index should export all stuff is needed.
You do not want to jump thru files to know that their implemented somewhere else.

cloud/wfm/src/executor/ExecutorImpl.ts
```diff
@@ -0,0 +1,37 @@
...
+export default class ExecutorImpl implements Executor {
```
wtrocki: Missing comments. They are valuable information. 
We should explain this model this as if not now then probably never :P

cloud/wfm/src/executor/ExecutorImpl.ts
```diff
@@ -0,0 +1,37 @@
...
+  public instance: ProcessInstance;
+  constructor(public process: Process, public instanceRepository: InstanceRepository) {
+    this.instance = this.process.createInstance();
+  }
+  public start() {
+    this.runCurrentTask();
+  }
+  protected onTaskDone(e: TaskEventData<Task>) {
```
wtrocki, line 23: Can we have this and possibly other events as part of the `TaskEventListener` interface? 
This interface can have state handlers like onTaskFinished onTaskStarted etc.
Thats good pattern when users will want to listen to events. Executor can both issue events and also provide default listeners. WDYT?

ps `onTaskDone` sounds strange. Are you ok with `onTaskFinished`?

cloud/wfm/src/executor/ExecutorImpl.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (e.currentStatus === TaskStatus.DONE) {
+      Promise.resolve(this.instance.next())
```
wtrocki: Self documenting code: `next` => `nextTask`

cloud/wfm/src/JsonSchema.ts
paolobueno: This is the typedef for an actual jsonschema object, so it's structure is at least somewhat validated during compile time:
```
import JsonSchema from './JsonSchema';
const evenNumberSchema: JsonSchema = {
  type: 'number',
  multipleOf: 2
};
```

The lib I picked for actually working with the schemas was [ajv](https://github.com/epoberezkin/ajv)

cloud/wfm/src/executor/ExecutorImpl.ts
```diff
@@ -0,0 +1,37 @@
...
+        .then(() => this.saveInstance())
+        .then(() => this.runCurrentTask());
+    }
+  }
+  protected saveInstance() {
+    return this.instanceRepository.save(this.instance);
```
wtrocki: Single responsiblity - should instance really have method to save itself?

cloud/wfm/src/executor/ExecutorImpl.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+  protected runCurrentTask() {
+    this.instance.getCurrentTask()
+    .then(t => {
+      // currentTask is undefined after all are done
```
wtrocki: Should this be normal case?  I think that getCurrentTask should be never undefined as it's not predictable. 
Last task that has status finished will be the best to mark entire ProcessInstance as finished.

cloud/wfm/src/executor/index.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Triggers instantiation and execution of a {@link ProcessInstance}
+ */
+export interface Executor {
+  process: Process;
+  instance?: ProcessInstance;
+  start(): void;
+}
+
+import ExecutorImpl, {InstanceRepository} from './ExecutorImpl';
+export {InstanceRepository};
```
wtrocki: Same as above.

cloud/wfm/src/index.ts
```diff
@@ -0,0 +1,19 @@
+export * from './executor';
```
wtrocki: As we agreed this files should start with Uppercase letter. All other PR's reflect this

cloud/wfm/src/index.ts
```diff
@@ -0,0 +1,19 @@
...
+export * from './process';
+export * from './process-instance';
+export * from './result';
+export * from './task';
+
+import ExecutorImpl from './executor';
+import ProcessImpl from './process';
+import ProcessInstanceImpl from './process-instance';
```
wtrocki: ProcesInstance as file name

cloud/wfm/src/index.ts
```diff
@@ -0,0 +1,19 @@
...
+import UrlResult from './result';
+import TaskImpl from './task';
+
+export {
```
wtrocki: @austincunningham @JameelB See this as good pattern to export multiple entires.

cloud/wfm/src/process-instance/InstanceImpl.ts
```diff
@@ -0,0 +1,61 @@
...
+export default class InstanceImpl extends EventEmitter implements ProcessInstance {
+  public id: string;
+  public assigneeId: string;
+  public processId: string;
+  public currentTask: Task;
+  public tasks: Task[];
+
+  protected currentTaskIdx: number;
+
+  // TODO: add repository for data storage
```
wtrocki: It's not clear for me how to do it here. Maybe we can create empty interface and move this TODO there?

cloud/wfm/src/task/index.ts
```diff
@@ -0,0 +1,91 @@
...
+   * Indicates that the Task has had its execution paused by a resolvable issue and can be resumed
+   */
+  'BLOCKED' = 400,
+  /**
+   * Indicates that the Task has finished in an unpredicted and unrecoverable state
+   */
+  'ERROR' = 500
+}
+
+type events = 'statusChange';
```
wtrocki: 1) Should this be array? 
2) We using this as string in multiple places. Can we have constant

cloud/wfm/src/executor/ExecutorImpl.ts
```diff
@@ -0,0 +1,37 @@
...
+      if (!t) {
+        return;
+      }
+      return t.on('statusChange', this.onTaskDone.bind(this)).run();
```
wtrocki: Maybe onTaskDone name should be more generic then and we may have different handlers inside?

cloud/wfm/src/process-instance/InstanceImpl.ts
```diff
@@ -0,0 +1,61 @@
...
+    const e: InstanceTaskEventData<this> = {
+      instance: this,
+      task: this.currentTask
+    };
+    this.emit('taskChange', e);
+    return Promise.resolve(this.currentTask);
+  }
+
+  public getCurrentTask() {
+    return Promise.resolve(this.currentTask);
```
wtrocki: [Antipattern] Promises for synchronous code. Why we need it? Interface compatibility?

cloud/wfm/src/process-instance/InstanceImpl.ts
```diff
@@ -0,0 +1,61 @@
...
+    return this.id;
+  }
+
+  public next() {
+    this.currentTask = this.tasks[++this.currentTaskIdx];
+    const e: InstanceTaskEventData<this> = {
+      instance: this,
+      task: this.currentTask
+    };
+    this.emit('taskChange', e);
```
wtrocki: This type of even seems to be unlisted. We have `statusChange` in events. 
Let's make set of constants that will list them (at least internal ones).
I know that this gives us a lots of flexibility but it's also really hard to read for developers.
When using constants we can track usages instead of classic text search.

cloud/wfm/src/process-instance/InstanceImpl.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  public getCurrentTask() {
+    return Promise.resolve(this.currentTask);
+  }
+
+  public getTasks() {
+    return Promise.resolve(this.tasks);
+  }
+
+  protected checkDone() {
```
wtrocki: 1) Missing documentation (method has complex return etc.)
2) This method can be generic and check any status. Check status can just accept one of the TaskStatuses

cloud/wfm/src/result/UrlResult.ts
paolobueno: Actually I think this one is very flexible, since most results could be represented by an URL.

Moving forward I see we needing things like `AppFormsResult`, `ImageResult` (with an accompanying frontend component rendering pictures on the mobile app), and so on, but most of those would still be able to be represented via a link the user could follow to download the resource from a simple http endpoint.

cloud/wfm/src/process-instance/index.ts
```diff
@@ -0,0 +1,43 @@
+import * as Promise from 'bluebird';
+import {EventEmitter} from 'eventemitter3';
+import * as _ from 'lodash';
+import {Process} from '../process';
+import {Task} from '../task';
+
```
wtrocki: Missing documentation. Every important!

cloud/wfm/src/process-instance/index.ts
```diff
@@ -0,0 +1,43 @@
...
+export interface InstanceTaskEventData<T extends ProcessInstance>  {
+  instance: T;
+  task: Task;
+}
+
```
wtrocki: Missing documentation. Every important!

cloud/wfm/src/process-instance/index.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * The executable instance of a {@link Process}
+ */
+export interface ProcessInstance extends EventEmitter {
+  assigneeId: string;
+
+  getTasks(): Promise<Task[]>;
+  getCurrentTask(): Promise<Task>;
```
wtrocki: Missing documentation.

cloud/wfm/src/process-instance/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  getId(): string;
+
+  /**
+   * Event emitted when the current active task changes to the next one
+   */
+  on(event: 'taskChange', handler: (e: InstanceTaskEventData<this>) => any): this;
+  /**
+   * Event emitted when all {@link Task}s achieve the done state
+   */
+  on(event: 'done', handler: (e: InstanceEventData<this>) => any): this;
```
wtrocki: Replace events with constants. 
It's just me thinking - Can we replace this events with just simple Listener pattern. For example:
https://docs.oracle.com/javase/tutorial/uiswing/events/intro.html

I'm not against events at all :) I just think that their are hard to document and debug later.

cloud/wfm/src/process/index.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Definition holder for a linear set of {@link Task}s
+ * Intended to be instantiated as a {@link ProcessInstance} in order to be executed
+ */
+export interface Process {
+  /** Unique identifier for this {@link Process} */
+  id: string;
+  /** Description for UI */
+  displayName: string;
+  tasks: Task[];
```
wtrocki: [Optional] Some implementation I saw store tasks in some additional container so different orders and flows can be applied later by providing separate implementation. 

Trick here is to not relay on the existence of the field and fact that is array. We can have simple TasksContainer that will just wrap array and provide methods to interact with it. 
That's just idea feel free to ignore.

cloud/wfm/src/result/BooleanResult.ts
```diff
@@ -0,0 +1,11 @@
+import {Result} from './index';
+
+/**
+ * Represents a {@link Result} that can be represented via a boolean value
+ * i.e. pass/fail, approval/rejection, etc.
+ */
+export default class BooleanResult implements Result {
```
wtrocki: Results will be always objects. Something like:
```{
 isHotDog: true
}```

File: cloud/wfm/src/result/UrlResult.ts (hunk 19 above)
Comment (wtrocki) [inline cloud/wfm/src/result/UrlResult.ts]: Same as above let's provide more accurate example that will inherit type from actual step. Steps always define specific result and they have full control over this.

File: cloud/wfm/src/task/BaseTask.ts [hunk 44]
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Base implementation for {@link Task}s
+ * Derived classes are expected to implement their run {@link #run} methods
+ */
+class BaseTask extends EventEmitter implements Task {
+  public result: Result;
+  /**
+   * Storage field for the `status` property
+   */
+  protected _status: TaskStatus | number = TaskStatus.PENDING;
```
wtrocki, line 14: avoid `_` as property name

`TaskStatus | number` - so custom statuses will be just numbers our customers will map by themselves? 
That's really clever :P

cloud/wfm/src/task/BaseTask.ts
```diff
@@ -0,0 +1,64 @@
...
+  protected options: object;
+
+  /**
+   * Setter for the status property
+   * Emits a statusChange event when set
+   */
+  public set status(to: TaskStatus | number) {
+    if (to === this._status) {
+      return;
```
wtrocki: This type of code can mask possible problems in developer code.
Worth to use logger here once it will be available ;)

cloud/wfm/src/executor/ExecutorImpl.ts
paolobueno: `onTaskDone` here is just an implementation detail for the executor for it to track the state of the instance's tasks, hence `protected`. Maybe we can set it up to forward the events, but separately.

I used 'done' because that's the name of the state, do you think we should change it?

Going to try out the `TaskEventListener` idea :+1: .

cloud/wfm/src/task/index.ts
```diff
@@ -0,0 +1,91 @@
...
+ */
+export interface Task {
+  status: TaskStatus | number;
+  result?: Result;
+  /**
+   * Set of runtime configuration options
+   * This is intended to be rendered as a `<form>` in a front-end application
+   * for instance by utilizing http://schemaform.io/
+   */
+  getOptionsSchema: () => JsonSchema;
```
wtrocki: So we want to use schema for our demo app tasks? Maybe we can have separate type of task that will have it.
Just thinking if that's optional

wtrocki (commented):
Minor comments. Missing documentation (README and global docs) and integration folder as example how to use this module.

cloud/wfm/src/result/UrlResult.ts
wtrocki: Sadly everything needs to be offline enabled. We cannot relay on any urls. I totally get you here, but I do think that we just need only object result for the moment to avoid confusion.

cloud/wfm/src/task/index.ts
paolobueno: This was an alias for more than one event: `'statusChange' | 'done'`, I removed the other one for now since it was just the special case where the new status was 'done'.

-1 from me for constants. I don't think they are very idiomatic in JS, it would be like having to do 
```javascript
$('#my-button').on(HTMLButtonElement.events.CLICK, e => ...);
```

I think that's because there are no true constants in JS, `events.CLICK === events['CLICK'] === 'click'`, so in the end you still need to know the right string, lol

FWIW vscode recognizes it and can do a global rename of this as a symbol

cloud/wfm/src/JsonSchema.ts
wtrocki: Did you meant runtime? 
IMHO validation is separate concern that we can provide this as extension to Step definition framework. I'm still not convinced why we need that file :)

The way I see it is that actual step developers can decide to validate their data before sending it as result. That's the best place. We can provide frameworks for that or even generate it like here: https://www.npmjs.com/package/typescript-json-schema

cloud/wfm/src/process-instance/index.ts
paolobueno: I took a better look into this, and it's also a -1 from me. TS is actually quite good for documenting events. `instance.on('something-else')` is a compile error

Intellisense support is also good:

image

Bear in mind that java uses listener interfaces because of language restrictions (functions aren't first class citizens, so they need interfaces + anonymous classes) and not because of OOP practices

cloud/wfm/src/task/BaseTask.ts
paolobueno: property name is `status` as per the interface, this is the storage variable for it, we might need to decide on a naming scheme.

Compiled js relies on [`Object.defineProperty`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty), from the compatibility notes I think it's a feature we can use.

cloud/wfm/src/result/BooleanResult.ts
paolobueno: I want to spike results better in [926](https://issues.jboss.org/browse/RAINCATCH-926), so could we merge these as-is for now?

We can also validate how we can integrate a new paradigm for results when doing [928](https://issues.jboss.org/browse/RAINCATCH-927) and [929](https://issues.jboss.org/browse/RAINCATCH-929)

cloud/wfm/src/JsonSchema.ts
paolobueno: The `JsonSchema.ts` file for compile-time checking and ajv for runtime validation.

I was thinking that by requiring a schema for configuring steps we could provide more value out of the box:

- generating forms for configuring a Task on the portal app
- validating the data coming from the configuration form
- in general provide a pattern for how tasks should be configured from the ui

We can defer this for later I guess, always a tradeoff between the base interfaces having more features and being 'customizable'.

paolobueno:
@wtrocki re-review when you have the chance, please!

I've changed istanbul's `check-coverage` to false for now so the build turns green, instead of reducing the values. I think we can reach our coverage goals in the follow up issues for wfm.

cloud/wfm/src/process-instance/index.ts
wtrocki: Awesome! I will need to play around with the actual usages to get my head around this.
Maybe as part of the exercise I can build integration tests? WDYT?

wtrocki:
@paolobueno I will build integration tests suite and README and PR against this branch. Then we will merge that.

wtrocki:
@paolobueno Tried to implement integration tests, but getting a lot of problems. 
I fixed some exports problems, but also getting couple of errors when actually trying to run this.
Let's sync on this once you are online.

cloud/wfm/src/executor/ExecutorImpl.ts
```diff
@@ -0,0 +1,57 @@
...
+    this.instance.getCurrentTask()
+    .then(task => {
+      // currentTask is undefined after all are done
+      if (!task) {
+        return;
+      }
+      return task
+        // TODO: this assumes the Task object will be kept in memory forever
+        // but human-executed tasks can take very long to complete
+        // we might have to change this to an out-of-process message bus
```
wtrocki: This code should execute on the client.
Everything will be sync based. Storage should be represented by repositories. 
Tasks will be stored on the client. Am I missing anything?
Worth to talk about this to avoid confusion.

cloud/wfm/src/index.ts
```diff
@@ -0,0 +1,33 @@
+// Interfaces
+import Executor from './executor/Executor';
+import ProcessInstance from './process-instance/ProcessInstance';
+import Process from './process/Process';
+import Result from './result/Result';
+import Task from './task/Task';
+
+// Implementations
+import ExecutorImpl from './executor/Executor';
```
wtrocki: [Optional] `ExecutorImpl` - BaseExecutor or DefaultExecutor or ProcessExecutor - remember we were saying to avoid Impl or Interface.

cloud/wfm/src/executor/ExecutorRepository.ts
```diff
@@ -0,0 +1,6 @@
+import ProcessInstance from '../process-instance/ProcessInstance';
+interface ExecutorRepository {
+  saveInstance(instance: ProcessInstance): Promise<ProcessInstance>;
```
wtrocki: Getting error that promise is not defined. Is there any global for this. How this works?

cloud/wfm/src/process/ProcessImpl.ts
```diff
@@ -0,0 +1,20 @@
+import Instance from '../process-instance/ProcessInstanceImpl';
+import Task from '../task/Task';
+import Process from './Process';
+
+export default class ProcessImpl implements Process {
+  public tasks: Task[];
```
wtrocki: It's not clear here how tasks will appear here.

cloud/wfm/src/task/BaseTask.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  /**
+   * Start the execution.
+   * This method is expected to be overridden in specific implementations
+   *
+   * This implementation simply sets the status to {@link TaskStatus#done}
+   */
+  public run() {
+    this.status = TaskStatus.DONE;
```
wtrocki, line 52: Should users deal with task status? Also method needs to be async and return promise to cover some use cases.

wtrocki, line 52: TaskStatus is not exposed

cloud/wfm/src/executor/ExecutorRepository.ts
paolobueno: ah, missing the bluebird import here, though it would fall back to the native promise

cloud/wfm/src/task/BaseTask.ts
paolobueno: I don't think a promise is enough, since a lot of Tasks can take hours to change state.

The idea was for events to capture the long-term async nature of these, but we'll probably have a better notion for this interface once we have a few custom Tasks

cloud/wfm/src/process-instance/ProcessInstance.ts
```diff
@@ -0,0 +1,41 @@
+import * as Promise from 'bluebird';
+import {EventEmitter} from 'eventemitter3';
+import * as _ from 'lodash';
+import Process from '../process/Process';
```
austincunningham: these don't appear to be used 
``` typescript
import * as _ from 'lodash';
import Process from '../process/Process';
```

cloud/wfm/src/task/BaseTask.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  /**
+   * Setter for the status property
+   * Emits a statusChange event when set
+   */
+  public set status(to: TaskStatus | number) {
+    if (to === this._status) {
+      // TODO: replace with logger object
+      // tslint:disable-next-line:max-line-length
+      console.warn('BaseTask#status setter: attempted status change to same status as current, no event will be emitted');
```
austincunningham, line 25: Logger is merged so can do this now.

cloud/wfm/src/index.ts
```diff
@@ -0,0 +1,33 @@
...
+import ProcessInstanceImpl from './process-instance/ProcessInstanceImpl';
+import ProcessImpl from './process/ProcessImpl';
+import BooleanResult from './result/BooleanResult';
```
austincunningham: Unused import 

``` typescript
import BooleanResult from './result/BooleanResult';
```

cloud/wfm/test/ProcessInstance/index.ts
```diff
@@ -0,0 +1,45 @@
+import * as assert from 'assert';
+import ProcessInstance, {InstanceEventData} from '../../src/process-instance/ProcessInstance';
+import InstanceImpl from '../../src/process-instance/ProcessInstanceImpl';
+import BaseTask from '../../src/task/BaseTask';
```
austincunningham: unused import

``` typescript
import InstanceImpl from '../../src/process-instance/ProcessInstanceImpl';
```

austincunningham (changes requested):
Needs integration implemented

austincunningham (dismissed):
I see wtrocki is doing integration tests, verified all tests passing and module starts.

cloud/wfm/src/executor/ExecutorImpl.ts
paolobueno: Makes sense, I'll remove this comment

cloud/wfm/src/task/BaseTask.ts
paolobueno: added to main module's exports

cloud/wfm/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Event-based tracking of Task progress
+
+Tasks and related entities emit events upon changing state, most of the regular progression should be between states in the [`TaskStatus` enumeration](./src/task/Task.ts), and custom states should fall under one of those as categories. See the [next section](#utilizing-custom-states) to learn how to utilize custom states.
+
+Through those events, both UI components and extra processing logic can hook into the life cycle of a Task in order to trigger other actions or update visible information for the user.
+
+### Utilizing custom states
+
+The [`TaskStatus` enumeration](./src/task/Task.ts) utilizes each hundredth number to represent a status that is expected to have semantic meaning for UI components and business logic.
```
wtrocki, line 43: `each hundredth number ` => numbers?

cloud/wfm/examples/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  ProcessInstanceImpl,
+  Result,
+  Task,
+  TaskStatus
+} from '../src';
+
+// Derive from BaseTask to implement one that deals with custom business logic
+class MyTask extends BaseTask {
+  public run() {
+    this.status = TaskStatus.ASSIGNED;
```
wtrocki, line 19: [Antipattern] Can we use setter method. It will allow us to do more than just set status.

wtrocki (approved):

cloud/wfm/examples/index.ts
paolobueno: This is already a setter method in basetask: https://github.com/feedhenry-raincatcher/raincatcher-core/pull/10/files/43dd2af2593788ed5af1b99f25a933c55e9c475e#diff-07e6fd89647c2a4e0697c91d131aa024R21

it's already taking care of publishing the event, and the unit tests for Task make sure other compatible implementations should do the same: https://github.com/feedhenry-raincatcher/raincatcher-core/blob/RAINCATCH-915-wfm/cloud/wfm/test/Task/index.ts#L22

PR #12, opened 2017-06-14
Add linting to travis

wtrocki:
## Description
Adding ts-lint to be run as travis build. 

## Progress
- [x] Added lint to travis

### DIFF
--- a/demo/server/src/index.ts
+++ b/demo/server/src/index.ts
@@ -1,9 +1,9 @@
 'use strict';
 /**
  * Module dependencies.
-*/
-import app from './app';
+ */
 import * as http from 'http';
+import app from './app';
 
 /**
  * Get port from environment and store in Express.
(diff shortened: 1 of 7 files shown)


paolobueno:
so from failed build, we need to add `tslint` and maybe `typescript` to devDependencies

paolobueno:
The npm lint command could also be changed to `tslint '*/*/src/**/*.ts' '*/*/test/**/*.ts'` (not sure if it supports two globs actually.)

Also it might be better to remove the `--fix` flag, because we want the job to error so the developer run --fix locally and then commit.

wtrocki:
@paolobueno 👍  Applied your comments and fixed current lint issues :D

demo/server/src/index.ts
```diff
@@ -40,6 +40,7 @@ function onError(error: any) {
  * Event listener for HTTP server "listening" event.
  */
 function onListening() {
-  var addr = server.address();
+  const addr = server.address();
+  // tslint:disable-next-line:no-console
```
JameelB: remove comment

PR #13, opened 2017-06-15
RAINCATCH-1014 - Fix local development issue due to compiling on prepublish

paolobueno:
## Motivation
The assumption that `ts-node` favors ts files over js ones when requiring a directory was wrong, so having js files compiled alongside the ts sources didn't solve the problem of not needing to recompile internal dependencies during local development.

## Description
Remove prepublish commands and add global build command to the repo root.

## Progress
- [x] remove prepublish from example/base
- [x] add global `npm run build`
- [x] add note to manually build the modules before publishing

ping @wtrocki


wtrocki (approved):
Verified locally.

wtrocki:
@feedhenry-raincatcher/rcdev After merging this change we will need to apply them to our modules that are in PR state now.

paolobueno:
so the 'migration instructions' are:

- Remove the prepublish npm script
- Remove the proxy `/index.js`
- Add `"main": "src/"` to package.json
- Optionally add `npm run clean` before the test command just to make sure.

PR #18, opened 2017-06-21
RAINCATCH-915 - integration tests

wtrocki:
## Motivation

Tried to verify module myself by writing integration tests. 
Minor adjustments made inside module itself. 
Added integration tests (we will need to improve them a little bit, after some of the additional comments will be resolved) 

ping @paolobueno

### DIFF
--- a/cloud/wfm/integration/index.ts
+++ b/cloud/wfm/integration/index.ts
@@ -1,0 +1,46 @@
+import * as Promise from 'bluebird';
+import {
+  BaseTask,
+  Executor,
+  ExecutorImpl,
+  ExecutorRepository,
+  Process,
+  ProcessImpl,
+  ProcessInstance
+} from '../src/index';
+
+// Create set of tasks you want to store
+class MyTask extends BaseTask {
+  public name: string;
+
+  constructor(name: string) {
+    super();
+    this.name = name;
+  }
+
+  public run() {
+    // tslint:disable-next-line:no-console
+    console.log('Executed task: ' + this.name);
+  }
+}
+
+// Create example process
+const exampleProcess: Process = new ProcessImpl('Example');
+exampleProcess.tasks = [new MyTask('StartProcessing'), new MyTask('CheckInventory')];
+
+// Store process instances.
+const repository: ExecutorRepository = {
+  saveInstance(instance: ProcessInstance): Promise<ProcessInstance> {
+    // Save data here
+    return Promise.resolve(instance);
+  }
+};
+
+const executor: Executor = new ExecutorImpl(exampleProcess, repository);
+
+// Actual processing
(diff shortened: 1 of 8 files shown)


wtrocki:
Getting error "Task list must be set before calling createInstance" - investigating.
I think that we may need to enforce that on constructor?

cloud/wfm/src/executor/Executor.ts
```diff
@@ -6,7 +6,7 @@ import Process from '../process/Process';
  *
  * Triggers instantiation and execution of a {@link ProcessInstance}
  */
-interface Executor {
+export interface Executor {
```
paolobueno, line 9: These are already the default exports. any reason for exporting twice per file?

```javascript
import Executor, {Executor as Executor2} from './Executor';
assert(Executor === Executor2);
```

cloud/wfm/integration/index.ts
```diff
@@ -0,0 +1,46 @@
...
+};
+
+const executor: Executor = new ExecutorImpl(exampleProcess, repository);
+
+// Actual processing
+// tslint:disable-next-line:no-console
+console.log('[Integration] Starting process');
+// TODO explanation this looks like execution is async here.
+executor.start();
+//
```
paolobueno, line 46: Ah, very nice skeleton. The unit tests for each entity already involves using the others, though.

I'm not sure the term 'integration' testing makes sense within a single module, even though there are many moving pieces.

I was thinking about having these along with sync + a web api

paolobueno:
@wtrocki about the `createInstance()` error yeah, maybe we could change the `ProcessImpl` constructor to take in the list of tasks already.

If you look at the unit tests the current setup is as follows:
https://github.com/feedhenry-raincatcher/raincatcher-core/blob/RAINCATCH-915-wfm/cloud/wfm/test/Process/index.ts#L11

cloud/wfm/src/executor/Executor.ts
wtrocki: As per typesript guideline default exports should be avoided. Always export stuff normally to allow * import. defaults aren't imported with star interface. See other PRs

cloud/wfm/integration/index.ts
wtrocki: That was something that was asked to be done.
In fact for this module we will need to do tests with couple other modules.. What name you propose?

wtrocki:
Good idea with moving this to constructor. Setting tasks is not intuitive. I'm also not sure how to continue this. Executor has no methods to move things around etc.

cloud/wfm/src/executor/Executor.ts
paolobueno: I see, we can merge this as is and remove all the default exports instead :+1:

PR #19, opened 2017-06-21
RAINCATCH-923 - DataStore helpers

wtrocki:
## Motivation

WIP Provide place for experiments on datastore.

## Supported objects

- [x] Paging (will be reused in WebAPI)
- [x] Sorting
- [ ] Query logic

## Idea

I think that we can have wfm-mongo, sec-mongo module that will implement all repositories and will include some interfaces from here. This one will include common elements.  This may sound messy but I couldn't really get anything reasonable that will be not limited. Providers should retrieve connection from top level application.

After building extensible support for queries I realized that we do not need that as users will just pass actual values to the repository methods and methods descriptions will need to explain clearly what we should query. For example:

```typescript
listUsersByNameAndManagerId(page:PageRequest, name:string, managerId:string) => PageResult<User>
```

Method name needs point to query specifics. For example:
```typescript
listUsersByNameOrManagerId(page:PageRequest,sort:SortOrder, name:string, managerId:string) => PageResult<User>
```
alternative:

``` typescript
interface Query {
  page : PageRequest, 
  name : string, 
  managerId : string
}
listUsersByNameAndManagerId(data: Query) => PageResult<User>
```

This is heavily inspired by SpringData ORM.

This can possibly make our methods name little bit long, but this will provide a lot of flexibility and will not require us to build any extensive data storage solution.

With this we really will need to have a lots of documentation for this process and team awareness.

Link entire team for comments: @feedhenry-raincatcher/rcdev

### DIFF
--- a/cloud/datastore/src/providers/MongoRepository.ts
+++ b/cloud/datastore/src/providers/MongoRepository.ts
@@ -1,0 +1,17 @@
+var MongoClient = require('mongodb').MongoClient
+  , assert = require('assert');
+
+// Idea
+// We can have wfm-mongo module that will implement all repositories and will include some interfaces from here.
+// Providers should retrieve connection from top level application.
+
+export class MongoProvider {
+  constructor(url:string){
+    var url = 'mongodb://localhost:27017/myproject';
+    MongoClient.connect(url, function(err, db) {
+      assert.equal(null, err);
+    });
+  }
+
+}
+
--- a/cloud/datastore/src/sorting/Sort.ts
+++ b/cloud/datastore/src/sorting/Sort.ts
@@ -1,0 +1,19 @@
+/**
+ * Sorting directions
+ */
+export enum DIRECTION {
+  ASC,
+  DESC
+}
+
+
+/**
+ * Interface used to monitor sorting requirements
+ */
+export interface SortOrder {
+  // List of fields to sort results
+  sortData: string[]
+  // Sort direction
+  direction: DIRECTION
+
+}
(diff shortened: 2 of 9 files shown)


cloud/datastore/src/providers/MongoRepository.ts
```diff
@@ -0,0 +1,17 @@
+var MongoClient = require('mongodb').MongoClient
```
wtrocki, line 1: This should be removed. Was just part of the spike

cloud/datastore/src/sorting/Sort.ts
```diff
@@ -0,0 +1,19 @@
...
+
+
+/**
+ * Interface used to monitor sorting requirements
+ */
+export interface SortOrder {
+  // List of fields to sort results
+  sortData: string[]
+  // Sort direction
+  direction: DIRECTION
```
paolobueno, line 17: probaby a type `'ASC' | 'DESC'` fits better here than an enum, the ts implementation is quite heavyweight.

https://www.typescriptlang.org/docs/handbook/enums.html (see 'compiled to:')

wtrocki: Cool, but what  you think about idea above?

paolobueno:
:+1: for the proposals in the `Idea` section. Smaller data-accessing methods that are very explicit in what they expect as a return are intrinsically well-documented and easier for users to understand.

The pagination helpers also seem a nice addition, from the client-side they could be used to access a [cursor-based api](https://dev.twitter.com/overview/api/cursoring), also [this article](https://developers.facebook.com/docs/graph-api/using-graph-api/v2.4#paging)

wtrocki:
Closing as stale PR. We need to rethink this idea again.

PR #25, opened 2017-06-27
WFM vehicle-inspection integration example

paolobueno:
## Motivation
Add a full example of the usage of the wfm module, with a back-end targetting the 

## Progress
- [x] Draft based on example from `@raincatcher/wfm`
- [x] REST api for creating a Process and configuring its Tasks (for the Portal app)
- [ ] Sync-based api for updating a ProcessInstance and its Tasks as they're executed (for the Mobile app)
- [x] Split index.ts into files for each entity
- [ ] Integration tests with mongo via supertest/curl + bare sync client as a way to showcase the example
- [ ] Complete README

ping @wtrocki


examples/vehicle-inspection/src/index.ts
```diff
@@ -0,0 +1,91 @@
...
+  private instance: ProcessInstance;
+  public saveInstance(instance: ProcessInstance) {
+    this.instance = instance;
+    return Promise.resolve(this.instance);
+  }
+}
+
+const repository = new InMemoryExecutorRepository();
+
+const executor: Executor = new ExecutorImpl(inspectionProcess, repository);
```
wtrocki: executor has very little responsibility now now as it's just have start method. We may drop that completely.

examples/vehicle-inspection/src/index.ts
```diff
@@ -0,0 +1,91 @@
...
+      throw new Error('This Task must be assigned to a user before its form is filled');
+    }
+
+    if (this.options.requirePictures) {
+      if (_.isEmpty(data.pictures)) {
+        throw new Error('Picture data for the Vehicle Inspection is required');
+      }
+    }
+    this.result = new VehicleInspectionFormResult(data);
+    this.status = TaskStatus.DONE;
```
wtrocki: IMHO: This statement is really misleading for someone who reads that. 
`updateStatus` method will be much cleaner and easier to understand.
Readability is the key.

wtrocki:
@paolobueno If we started working on this let's pick ticket. 
This is actually really valid example and in fact work we need to do. 
It's best time to spike it. 

Couple comments for this:
1) I think we should use  this repo for this work 
https://github.com/feedhenry-raincatcher/raincatcher-angularjs

2) If we integrate this work with angular directive we will be able to show this (that will be awesome)
https://github.com/feedhenry-raincatcher/raincatcher-vehicle-inspection

3) This example can be still duplicated in this repository but let's move this to be official example for WFM without separate folder. Let's keep repository organized and clean.

paolobueno:
@wtrocki I was thinking of moving this work to be under https://issues.jboss.org/browse/RAINCATCH-928 if we bring it into the sprint, WDYT?

We can integrate with the changes in the `-angularjs` module, but since this would need to depend on the unpublished wfm and sync modules I don't think there's a way to add them to that repository.

examples/vehicle-inspection/src/index.ts
paolobueno: agreed, going to move from setter to an explicit method

examples/vehicle-inspection/src/index.ts
paolobueno: It definitely seems so. It's role is to avoid having manipulation methods on the other entities and make it so they're mostly data bags with some event emitting.

We might decide to merge it into the `ProcessInstance`

examples/vehicle-inspection/src/index.ts
paolobueno: the difficult part is just on the interface side, since you can't have a readonly member:
```typescript
interface Task {
  status:number; // no guarantee that users will always use updateStatus
  updateStatus(to: number);
  getStatus(): number; // this looks a bit ugly
}
```

but you can implement a member via a pair of getter+setter functions.

wtrocki: We may drop member and go with methods only. WDYT? We have base implementation for that anyway and users will never really build using only interface? Am I right?

paolobueno: :+1: That's probably the way to go, remove the member from the interface.

It seems TS has a `readonly` modifier that can also be present in interfaces, but I need to look into how it gets compiled, it's probably only a compile-time check.

lerna.json
```diff
@@ -3,6 +3,7 @@
   "packages": [
     "cloud/*",
     "client/*",
+    "examples/*",
```
wtrocki: Let's move this to integration as it's really good integration tests base.

cloud/wfm/src/index.ts
```diff
@@ -15,3 +15,6 @@ export {BaseTask} from './task/BaseTask';
 
 // Repositories
 export {ExecutorRepository} from './executor/ExecutorRepository';
```
wtrocki: `XRepository` classes are for fetching data. Users can map this to mongo requests etc.

cloud/wfm/src/process-instance/ProcessInstanceImpl.ts
```diff
@@ -47,7 +47,7 @@ export class ProcessInstanceImpl extends EventEmitter implements ProcessInstance
   }
```
wtrocki: That's ex Workflow. We adjusted names to BPMN process

cloud/wfm/src/task/BaseTask.ts
```diff
@@ -12,22 +12,22 @@ export class BaseTask extends EventEmitter implements Task {
    * Storage field for the `status` property
```
wtrocki: Task is ex `Step`

wtrocki:
ping @twnolan

integration/vehicle-inspection/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Sequential use cases
+
+1. Create a new Process
+2. Add a `VehicleInspectionTask` to the new Process
+    * Configure it via a Form generated from the JsonSchema of the custom `Task`
+3. Create a ProcessInstance via a Executor and assign it to a User
+4. As the assigned-to User, on the mobile app, start the execution of the `ProcessInstance`
+5. Finish the execution of the `VehicleInspectionTask`, moving it to the `TaskStatus.DONE` state
+6. Visualize the results of the `VehicleInspectionTask`
```
paolobueno, line 15: @wtrocki in the actual usage, steps 4 and 5 would be done through sync from the 'mobile app', but for simplicity in this demo should we go for a REST api instead?

paolobueno:
@twnolan feel free to pick up this branch already, always available to clear things up.

I'll probably add some better info to the README and add some jsdocs around the stuff that's in here, but hopefully no merge conflicts.

integration/vehicle-inspection/README.md
wtrocki: I'm actually going to have this implemented in VehicleInspection module I'm building.
I'm not sure how we going to do it now. Going to update this PR as soon I will get this changes back.

paolobueno: I was thinking that as an other option. Let's get this one merged so we can move things around in a follow up for the other JIRA.

paolobueno:
note: rebased on #31

client/wfm/src/process-instance/ProcessInstanceRepository.ts
```diff
@@ -0,0 +1,8 @@
+import {ProcessInstance} from './ProcessInstance';
+export interface ProcessInstanceRepository {
+  getAll(): Promise<ProcessInstance[]>;
```
wtrocki: getAllForUser and getAll should use paggination. 
We should add todo to indicate that this needs to be paginated.

paolobueno: I think we should include the pagination stuff in the JIRAs for the store. +1 on adding the TODO comment though.

client/wfm/src/task/TaskRepository.ts
```diff
@@ -0,0 +1,5 @@
+import {Task} from './Task';
+export interface TaskRepository {
+  getAll(): Promise<Task[]>;
```
wtrocki: Should this be getAllFor Process or processInstance?

client/wfm/src/task/TaskRepository.ts
```diff
@@ -0,0 +1,5 @@
...
+  createBatch(tasks: Task[]): Promise<Task[]>;
```
wtrocki: Without documentation it's really unclear what that is doing.

paolobueno: going to add jsdocs, probably rename this method specifically.

client/wfm/src/process-instance/ProcessInstanceRepository.ts
```diff
@@ -0,0 +1,8 @@
...
+  getById(id: string): Promise<ProcessInstance>;
+  getByProcessId(id: string): Promise<ProcessInstance[]>;
```
wtrocki: IS that used? IMHO we may never use this actual method as it will be overkill.
We always looking for filtered ProcessInstances (probably per assignee etc.)

client/wfm/src/task/TaskRepository.ts
paolobueno: Probably, doesn't make sense to grab *all* Tasks in the world, not using it anywhere in the example atm, but just thought maybe all repositories would have CRUD**L**

integration/vehicle-inspection/src/index.ts
```diff
@@ -0,0 +1,15 @@
+import {ConsoleLogger} from '@raincatcher/logger';
+import {
+  InMemoryProcessInstanceRepository,
+  InMemoryProcessRepository,
+  InMemoryTaskRepository
+} from './repositories';
+import {Server} from './server';
+
+const server = new Server(
```
wtrocki: ServerName may be misleading. What is server here? WebServer?

client/wfm/src/process-instance/ProcessInstanceRepository.ts
paolobueno: Used [here](https://github.com/feedhenry-raincatcher/raincatcher-core/blob/vehicle-inspection-example/integration/vehicle-inspection/src/server/index.ts#L62-L66), but mostly for convenience/completeness. I think the POST method is more important there.

We can also move it to `POST /[process-]instances`, but we'd probably have to include the process id in the payload.

integration/vehicle-inspection/src/server/index.ts
```diff
@@ -0,0 +1,116 @@
...
+    // TODO: Add new security module
+  }
+
+  protected setupExpressRoutes() {
+    this.app.use('/processes', this.buildProcessRouter());
+    this.app.use('/instances', this.buildProcessInstanceRouter());
+    this.app.use('/tasks', this.buildTaskRouter());
+  }
+
+  protected buildProcessRouter() {
```
wtrocki: Awesome! Is that for portal usage? Looks like you nailed that already. Not sure if that's actually serving purpose of this integration tests but we can move that out to actual cloud app or separate module!

integration/vehicle-inspection/src/server/index.ts
```diff
@@ -0,0 +1,116 @@
...
+    }));
+
+    return taskRouter;
+  }
+
+  protected createProcess(formData: any) {
+    const tasks: Task[] = map<any, Task>(formData.tasks, taskData => {
+      let task: Task;
+      // We can move this to a more robust dictionary-based registerTaskFactory(factory: (data) => Task)
+      switch (taskData.type) {
```
wtrocki: IMHO. That looks terrible. We will need to figure out how to do it properly. 
Let's not bother for now.

integration/vehicle-inspection/src/server/index.ts
paolobueno: Didn't think that far, after we merge this we can definitely take what's good and reuse :)

integration/vehicle-inspection/src/server/index.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    return processInstanceRouter;
+  }
+
+  protected buildTaskRouter() {
+    const taskRouter = express.Router();
+    // Looks like the options schema would be a better fit for a static property,
+    // but it's complicated to enforce it on typescript:
+    // https://github.com/Microsoft/TypeScript/issues/13462#issuecomment-272669082
+    const vehicleInspectionSchema = new VehicleInspectionTask().getOptionsSchema();
```
wtrocki: I think that will be overkill for the moment, but there is absolutely no problem to have this in this particular PR.We can decide how we going to use it (or if) later.

wtrocki (dismissed):
Minor comments added, but overall I think all of that can be improved over the time. 
This set of integration stuff will need to be long running task and we should update as we go

wtrocki (approved):

PR #29, opened 2017-06-29
Logger unit tests

spriadka:
## Description
A few sentences describing the overall goals of the pull request's commits.

## Progress
- [ ]
- [ ] Unit tests
- [ ] Documentation


spriadka:
@wtrocki Can you look at this?

.vscode/launch.json
```diff
@@ -1,20 +1,17 @@
-{
-  // Use IntelliSense to learn about possible Node.js debug attributes.
```
wtrocki, line 2: That should be removed

cloud/datasync/package.json
```diff
@@ -35,10 +35,12 @@
     "@types/proxyquire": "^1.3.27",
     "del-cli": "^1.0.0",
     "mocha": "^3.4.2",
+    "mocha-typescript": "^1.1.5",
     "nyc": "^11.0.1",
     "proxyquire": "^1.8.0",
     "source-map-support": "^0.4.15",
     "ts-node": "^3.0.4",
+    "typemoq": "^1.7.0",
```
wtrocki, line 43: We will need to add this to ./templates/base and makes sure that developers will also incorporate that change.

cloud/datasync/test/SyncServerTest.ts
```diff
@@ -74,3 +76,42 @@ describe('FeedHenry Sync Tests', function() {
...
+    let customDataSetHandler: DataSetHandler = new MyTestDataSetHandler();
+    return this.syncApi.setGlobalDataHandlers(customDataSetHandler);
+  }
+
+  @test
+  public testCallRegisterDataSetHandler(){
+    let options: SyncDataSetOptions = {
+        backendListTimeout: 20,
+        // tslint:disable-next-line:no-empty
+        hashFunction() {
```
wtrocki, line 107: This methods are optional so you do not need to provide them.

.vscode/tasks.json
```diff
@@ -0,0 +1,9 @@
+{
```
wtrocki: Remove this

.vscode/tasks.json
```diff
@@ -0,0 +1,9 @@
...
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "0.1.0",
```
wtrocki: add .vscode to gitignore

cloud/datasync/test/SyncServerTest.ts
```diff
@@ -74,3 +76,42 @@ describe('FeedHenry Sync Tests', function() {
     });
   });
 });
+
+@suite
```
wtrocki, line 80: I'm not sure if we want to use typemoq. We can introduce it in the places where it's needed, without migrating all other tests.

ping @paolobueno

cloud/datasync/package.json
```diff
@@ -35,10 +35,12 @@
     "@types/proxyquire": "^1.3.27",
     "del-cli": "^1.0.0",
     "mocha": "^3.4.2",
+    "mocha-typescript": "^1.1.5",
```
paolobueno, line 38: -1 for using this, I had investigated it before. It breaks the goal that our test suites be usable from alternative, plain js implementations.

See the [js template](https://github.com/feedhenry-raincatcher/raincatcher-core/tree/master/templates/javascript-usage) for an example of JS code using the compiled TS tests.

package.json
```diff
@@ -5,6 +5,7 @@
   "main": "index.js",
   "scripts": {
     "test": "lerna run test",
+    "clean": "lerna exec npm run clean",
```
paolobueno, line 8: this is best done through `lerna run clean`, it's almost the same but I think the logging output is better.

wtrocki:
My overall impression is that we should really contribute new tests here and initiate conversation with the team and community around what tooling we should use for testing.

cloud/datasync/test/SyncServerTest.ts
paolobueno: These annotations are actually mocha-typescript stuff. -1 from me from the other comment. It won't allow us to have a good support for our test suites to be used by non-TS code.

In addition it suggests a different naming that what we usually go for when using mocha's default BDD interface.
i.e. instead of `SyncServerTest#testConnect()` we would usually have `'SyncServer' ... 'should be able to connect...'`

paolobueno: As for typemoq it doesn't look like it's actually being used here. I think some of the other modules had success with sinon, I'd rather have it instead just because it's so much, much more popular, which makes it more familiar to community contributors.

cloud/datasync/package.json
paolobueno: more precisely [this line](https://github.com/feedhenry-raincatcher/raincatcher-core/blob/master/templates/javascript-usage/test/index.js#L9) calls into the TS test suite

cloud/datasync/test/SyncServerTest.ts
wtrocki: Agreeing on this. @feedhenry-raincatcher/rcdev opinions?

wtrocki:
@spriadka - Thanks for contribution. 
Let's initiate conversation with the team for this particular changes.
Closing for now as this will quickly get outdated.

PR #30, opened 2017-06-29
Fix sync client outdated ts options file

wtrocki:
## Motivation

Sync client was compiling source code to dist folder, which was not standard behavior for clients.
This change fixes configuration to standard behaviour


austincunningham (approved):
Agree as npm run clean doesn't check for dist

PR #33, opened 2017-07-03
Interface improvements for passport

wtrocki:
## Motivation

Improve passport interfaces. Clean up types and improve example application.

## Tasks

- [x] Refactoring for interface
- [x] Update documentation
- [x] Fix tests


cloud/passportauth/test/PassportAuthTest.ts
```diff
@@ -91,14 +89,14 @@ describe('Test Passport Auth', function() {
...
 
-    passport.authenticate('local')(mockReq as express.Request, mockRes as express.Response,
+    testSubject.authenticate('/')(mockReq as express.Request, mockRes as express.Response,
       mockNext as express.NextFunction);
 
     setTimeout(() => {
       sinon.assert.calledOnce(mockReq.logIn);
 
       done();
-    }, 500);
```
wtrocki: @JameelB We should avoid slowing down our tests.

cloud/passportauth/test/PassportAuthTest.ts
```diff
@@ -51,14 +49,14 @@ describe('Test Passport Auth', function() {
       user: null
     };
 
-    passport.authenticate('local')(mockReq as express.Request, mockRes as express.Response,
+    testSubject.authenticate('/')(mockReq as express.Request, mockRes as express.Response,
```
wtrocki, line 52: @JameelB Changed that as we should avoid testing third party components in unit tests.

wtrocki:
Generated typedocs documentation. I think that we should still improve our documentation over the time by adding more explanation about ideas and possible configuration options.

austincunningham (commented):
will have another look when the build is passing

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -111,10 +82,33 @@ export class PassportAuth implements Auth {
         }
         return res.redirect(this.loginRoute);
       }
-
-      return this.userSec.hasResourceRole(role) ? next() : res.status(401).send(new Error('Unauthorized'));
+      return this.userSec.hasResourceRole(role) ? next() : res.status(403).send();
```
wtrocki: @JameelB 403 status for lack of access to resource

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -17,91 +24,55 @@ export interface Auth {
...
-
-  public setupDeserializeUser(deserializeUser: (userSec: UserSec) =>
-      (user: any, cb: (error: Error|null, user: any) => any) => any) {
-    this.log.info('Setting up passport deserializeUser');
-    passport.deserializeUser(deserializeUser(this.userSec));
-  }
-
+  /**
+   * Function which checks if the user requesting access to the resource is authenticated and authorized to
+   * access the resource. Redirects to the login page if user is not authenticated or returns a status of 401
```
tomjackman: Small docs change - "status of 403"?

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -17,91 +24,55 @@ export interface Auth {
...
-   * Sets the deserializeUser function to be used by Passport
-   *
-   * @param deserializeUser {Function} - Function to be used by passport's deserializeUser
-   */
-  setupDeserializeUser(deserializeUser: (userSec: UserSec) =>
-    (user: any, cb: (error: Error|null, user: any) => any) => any): void;
-
-  /**
    * Function which checks if the user requesting access to the resource is authenticated and authorized to
    * access the resource. Redirects to the login page if user is not authenticated or returns a status of 401
```
tomjackman: Small docs change - "status of 403"?

wtrocki:
## Issues
- Passport is failing to run and build on node5
- Redirects will not work with secure cookies
- Disabled role base auth (separate ticket)

cloud/logger/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@raincatcher/logger",
-  "version": "0.0.1",
+  "version": "1.0.0",
```
paolobueno, line 3: :+1: We'll need to look again into the single versioning in the JIRA for setting up npm publish.

cloud/passportauth/example/index.ts
```diff
@@ -1,13 +1,14 @@
-import {BunyanLogger, Logger} from '@raincatcher/logger';
+import { BunyanLogger, Logger } from '@raincatcher/logger';
```
paolobueno, line 1: :+1: I tried to find a rule in tslint for this, looks like this one does it: https://palantir.github.io/tslint/rules/whitespace/ with `"check-module"`, but unfortunately doesn't seem to have an automatic fix.

cloud/passportauth/example/index.ts
```diff
@@ -33,26 +36,35 @@ app.use(bodyParser.urlencoded({
...
   res.sendFile(path.join(__dirname, 'public/login.html'));
 });
 
-app.post('/login',  passport.authenticate('local', { failureRedirect: '/login',
-  successReturnToOrRedirect: '/testUserEndpoint'}));
+app.post('/login', authService.authenticate('/'));
+
+app.use(function(err: any, req: express.Request, res: express.Response, next: any) {
+  // tslint:disable-next-line:no-console
+  console.log(err);
```
paolobueno: use logger module here?

cloud/passportauth/src/auth/DefaultStrategy.ts
```diff
@@ -0,0 +1,31 @@
...
+import { User } from '../user/User';
+
+/**
+ * Default strategy to be used by Passport's local strategy. If user credentials are valid, proceed to login,
+ * otherwise, reject it. Calls serialize user with user login id.
+ *
+ * @param userSec {UserSec}
+ * @returns {Function} - Returns the default strategy function to be used by passport
+ */
+export const DefaultLocalStrategy = (userSec: UserSecurityService) => {
```
paolobueno: File name DefaultStrategy.ts differs from the export's name

cloud/passportauth/src/auth/DefaultStrategy.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const log: Logger = new BunyanLogger({ name: 'Passport-Auth', level: 'error' });
+import { User } from '../user/User';
+
+/**
+ * Default strategy to be used by Passport's local strategy. If user credentials are valid, proceed to login,
+ * otherwise, reject it. Calls serialize user with user login id.
+ *
+ * @param userSec {UserSec}
+ * @returns {Function} - Returns the default strategy function to be used by passport
```
paolobueno: we should avoid using types in jsdocs, it's an extra point of maintenance. I think the doc generator we're using either ignores them in favor of the types given by the compiler or includes these as extra plaintext (which would be even worse).

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -17,104 +24,93 @@ export interface Auth {
...
    * if the user does not have the required role.
    *
    * @param role {string} - Role which the user needs in order to access this resource
    */
   protect(role?: string): void;
```
paolobueno: isn't the return an `express.Handler` instead of `void`?

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -17,104 +24,93 @@ export interface Auth {
...
-  constructor(protected readonly userSec: UserSec, loginRoute?: string) {
+  constructor(protected readonly userSec: UserSecurityService, loginRoute?: string) {
     this.loginRoute = loginRoute || '/login';
   }
 
+  /**
+   * Initializes an Express application to use passport and express-session
+   *
+   * @param app {express.Express} - An express application
+   * @param sessionOpts {SessionOptions} - Session options to be used by express-session
```
paolobueno: -1 for types in jsdocs here too

cloud/passportauth/src/auth/UserSerializer.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Default deserialize user function to be used by Passport. Attaches the user to
+ * req.user
+ *
+ * @returns {Function} - Returns the default deserialize user function
+ */
+export const DefaultDeserializeUser = (user: any, done: (error: Error | null, user: any) => any) => {
+  return done(null, user);
+};
```
paolobueno: Should we use upper case for non-constructors?

cloud/passportauth/test/PassportAuthTest.ts
paolobueno: maybe use `setImmediate()` instead of `setTimeout(_, 0)`

cloud/passportauth/test/PassportAuthTest.ts
```diff
@@ -51,14 +49,14 @@ describe('Test Passport Auth', function() {
...
       mockNext as express.NextFunction);
 
     setTimeout(() => {
       sinon.assert.notCalled(mockReq.logIn);
 
       done();
-    }, 500);
+    }, 0);
   });
```
austincunningham: Any point in setting a timeout if your setting it to zero

austincunningham (commented):
Tests failing

cloud/logger/package.json
wtrocki: I just do not want to waste time for bumps. If we release this sprint we can decide what version we need. For the moment best to keep versions the same

cloud/passportauth/example/index.ts
wtrocki: I have configured vscode and webstorm with formatter and making sure that our source code is properly formatted using that rules. Automatic fix is not needed as we will use IDE formatters that enforce this anyway. It's just good for checking if we not pushing unformatted code to master.

cloud/passportauth/src/auth/DefaultStrategy.ts
wtrocki: It's the latter. FYI @JameelB

austincunningham (dismissed):
verified

wtrocki:
## Outstanding Issues
- Improvements in typedocs and documentation needed
- Passport is failing to run and build on latest node 
- Redirects will not work with secure cookies ( @JameelB - let's talk about this and make sure we have jira for that issue)
- Disabled role base auth (separate ticket already in the jira)
- Login page has no info for failed attempts (already in the jira)

austincunningham (approved):
verified

cloud/passportauth/example/index.ts
paolobueno: yes, final checking via tslint in CI is the best we can have on top of editor support. How did you manage to configure vscode to add these? I couldn't find a way to

paolobueno: ah, found https://github.com/Microsoft/vscode/issues/18043

PR #36, opened 2017-07-07
RAINCATCH-1051

austincunningham:
## Motivation
User gets no feedback on incorrect login
## Description
Add a new route to handle incorrect login and add a 'invalid credentials error to login view'

## Progress
- [x] Added route and view
- [x] Unit tests
- [x] Documentation

## Additional Notes
screenshot from 2017-07-06 10-26-57

### DIFF
--- a/cloud/passportauth/src/auth/PassportAuth.ts
+++ b/cloud/passportauth/src/auth/PassportAuth.ts
@@ -103,20 +103,16 @@ export class PassportAuth implements Auth {
     });
   }
 
-  /*public authenticate(defaultRedirect?: any, ...params: any[]) {
-   return passport.authenticate(defaultRedirect);
-   }*/
-
   /**
    * Initialized passport configuration.
    * Method can be overriden to provide custom passport setup
    *
    * @param passport - passport.js instance
    */
-  protected setup(passport: passport.Passport) {
-    passport.use(new Strategy(defaultStrategy(this.userSec)));
-    passport.serializeUser(defaultSerializeUser);
-    passport.deserializeUser(defaultDeserializeUser);
+  protected setup(passportApi: passport.Passport) {
+    passportApi.use(new Strategy(defaultStrategy(this.userSec)));
+    passportApi.serializeUser(defaultSerializeUser);
+    passportApi.deserializeUser(defaultDeserializeUser);
   }
 }
 
--- a/cloud/passportauth/src/auth/PassportAuth.ts
+++ b/cloud/passportauth/src/auth/PassportAuth.ts
@@ -46,10 +46,12 @@ export interface Auth {
  */
 export class PassportAuth implements Auth {
   protected loginRoute: string;
-  private log: Logger = new BunyanLogger({ name: 'Passport-Auth', level: 'error' });
+  protected loginErrorRoute: string;
+  private log: Logger = new BunyanLogger({name: 'Passport-Auth', level: 'error'});
 
-  constructor(protected readonly userSec: UserSecurityService, loginRoute?: string) {
+  constructor(protected readonly userSec: UserSecurityService, loginRoute?: string, loginErrorRoute?: string) {
     this.loginRoute = loginRoute || '/login';
+    this.loginErrorRoute = loginErrorRoute || '/loginError';
   }
 
   /**
@@ -96,10 +98,14 @@ export class PassportAuth implements Auth {
    */
   public authenticate(defaultRedirect: string) {
     return passport.authenticate('local', {
-      failureRedirect: this.loginRoute,
+      failureRedirect: this.loginErrorRoute,
       successReturnToOrRedirect: defaultRedirect
     });
   }
+
+  /*public authenticate(defaultRedirect?: any, ...params: any[]) {
+   return passport.authenticate(defaultRedirect);
+   }*/
 
   /**
    * Initialized passport configuration.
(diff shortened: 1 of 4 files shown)


JameelB (dismissed):

JameelB (dismissed):

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -46,10 +46,12 @@ export interface Auth {
  */
 export class PassportAuth implements Auth {
   protected loginRoute: string;
-  private log: Logger = new BunyanLogger({ name: 'Passport-Auth', level: 'error' });
+  protected loginErrorRoute: string;
+  private log: Logger = new BunyanLogger({name: 'Passport-Auth', level: 'error'});
 
-  constructor(protected readonly userSec: UserSecurityService, loginRoute?: string) {
+  constructor(protected readonly userSec: UserSecurityService, loginRoute?: string, loginErrorRoute?: string) {
```
wtrocki: We do not need that here. Login route is used to mount endpoints and mark passport. For error route we should pass that directly to function

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -96,7 +98,7 @@ export class PassportAuth implements Auth {
    */
   public authenticate(defaultRedirect: string) {
     return passport.authenticate('local', {
-      failureRedirect: this.loginRoute,
+      failureRedirect: this.loginErrorRoute,
```
wtrocki: We can/should pass this here. As every authenticate may have different error route.

wtrocki (changes requested):
We need to move this parametere from constructor to method.

cloud/passportauth/src/auth/PassportAuth.ts
austincunningham: removed it from the constructor and moved it to the method

cloud/passportauth/src/auth/PassportAuth.ts
austincunningham: done

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -93,10 +93,11 @@ export class PassportAuth implements Auth {
    *
    * @param defaultRedirect - location to redirect after successful authentication
    *                          when login page was loaded directly (without redirect)
+   * @param loginErrorRoute - location to redirect after unsuccessful authentication
    */
-  public authenticate(defaultRedirect: string) {
+  public authenticate(defaultRedirect: string, loginError: string) {
```
wtrocki: loginErrorRoute:string?

We may add it as optional and default to login in case you do not want to redirect to any other page.
There is way for passport to redirect with some parameteres to same page.

wtrocki (dismissed):

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -38,15 +38,15 @@ export interface Auth {
    *
    * @param redirect - location to redirect after successful authentication
    */
-  authenticate(redirect: string): express.Handler;
+  authenticate(redirect: string, loginError: string): express.Handler;
```
paolobueno, line 41: Need jsdocs for second param, maybe rename to `errorRedirect`? I had to look for the usage to tell what it was, my initial guess was that it was an error message :joy:

paolobueno: maybe just use `failureRedirect` as passport does

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -107,10 +108,10 @@ export class PassportAuth implements Auth {
...
    * @param passport - passport.js instance
    */
-  protected setup(passport: passport.Passport) {
-    passport.use(new Strategy(defaultStrategy(this.userSec)));
-    passport.serializeUser(defaultSerializeUser);
-    passport.deserializeUser(defaultDeserializeUser);
+  protected setup(passportApi: passport.Passport) {
+    passportApi.use(new Strategy(defaultStrategy(this.userSec)));
+    passportApi.serializeUser(defaultSerializeUser);
+    passportApi.deserializeUser(defaultDeserializeUser);
```
paolobueno, line 114: Any reason for the rename? -1 for naming stuff *Api because it's redundant

wtrocki: It's my fault :) 
Basically it was shadowing exactly the same value available globally. 
What we trying to do here is to deliver this as parameter so users will not need to require original passport (we just preparing this method to be extended.

We just need to rename this to be something else than `passport`.

paolobueno: ah, my suggestion then is renaming the import instead of the parameter :+1:

paolobueno (changes requested):
couple tiny changes

cloud/passportauth/src/auth/PassportAuth.ts
austincunningham: tslint was complaining about shadow names, so will have to leave it as is

paolobueno: yeah, we can alias the `import * as passportApi from 'passport'` and keep the parameter with the prettier name, because the latter is expected to be seen from outside the module.

paolobueno (approved):

cloud/passportauth/src/auth/PassportAuth.ts
austincunningham: redirect route , updated the jsdocs :)

paolobueno: I think that change still needs to be applied to the interface too

PR #42, opened 2017-07-12
[Idea] Global logger example

wtrocki:
## Motivation

Showcase for team how we can mount logger to be available globally.

I just noticed that we instantiate logger for each class and that's not effective and produces a lots of boilerplate. Passing logger or having wrapper common.js module with single instance maybe the answer but this code will need to be duplicated across the all the modules. 

Proposed solution is simple and will do the job.

## Note

Code is not ready to be merged. Presenting this as idea that needs to be improved etc.

## Usage
```
import {logger} from '@raincatcher/austin's-cloud-logger' 
logger.info("Going to have break now");
```
💯 **Super ping**   @austincunningham 
🔬  **Micro ping** @witmicko

### DIFF
--- a/cloud/logger/src/index.ts
+++ b/cloud/logger/src/index.ts
@@ -4,3 +4,23 @@ import {ConsoleLogger} from './ConsoleLogger';
 export {Logger} from  './Logger';
 export {BunyanLogger} from  './BunyanLogger';
 export {ConsoleLogger} from './ConsoleLogger';
+
+
+import {Logger} from  './Logger';
+
+// Global logger that will be used in all applications
+export let logger: Logger; 
+
+/**
+* Set default logger to be used by all modules
+*/
+export function setGlobalLogger(defaultLogger: Logger){
+   logger = defaultLogger;
+}
+
+/**
+* Retrieve default logger
+*/
+export function getGlobalLogger(location?:string): Logger{
+   return logger;
+}


cloud/logger/src/index.ts
```diff
@@ -4,3 +4,23 @@ import {ConsoleLogger} from './ConsoleLogger';
...
+* Set default logger to be used by all modules
+*/
+export function setGlobalLogger(defaultLogger: Logger){
+   logger = defaultLogger;
+}
+
+/**
+* Retrieve default logger
+*/
+export function getGlobalLogger(location?:string): Logger{
```
wtrocki, line 24: We should wrap this probably in empty logger in case no logger is set. 
This will mean that users do not wish to see any logs from our modules.

paolobueno:
Could be a way to avoid constructor-based injection if consensus is that it pollutes the apis for other modules

wtrocki:
Constructor based injection works fine but it works only for specific class, but its hard to do that  for other files which have static methods etc.

austincunningham (commented):
Agree constructor based injections work but you will use logger in more that just classes. +1 for this approach

cloud/logger/src/index.ts
```diff
@@ -4,3 +4,23 @@ import {ConsoleLogger} from './ConsoleLogger';
 export {Logger} from  './Logger';
 export {BunyanLogger} from  './BunyanLogger';
 export {ConsoleLogger} from './ConsoleLogger';
+
+
+import {Logger} from  './Logger';
+
+// Global logger that will be used in all applications
+export let logger: Logger;
```
witmicko, line 12: shouldnt this be const since its global and intended to live alongside application?

austincunningham: const require initialization in this place, thats why this value needs to be set externally

wtrocki:
Idea accepted. Going to be implemented in separate PR.

PR #45, opened 2017-07-13
[Proposition] Rename security interface

wtrocki:
## Motivation

One element I missed in my previous PR was renaming top level interface. 
I think we should:

- define interface names more carefully
- avoid using implementation specifics for interfaces 
- avoid using shorthands for names (prop => Property, Auth => Authorization)

Note: This change is not ready to be merged. Just idea.
 
## Security interface

In this particular case Auth interface suggest that we may deal with Authentication or Authorization. Shorter name causing confusion etc. Our interface is doing both so best to find more top level name. My suggestion in changes.

## Community voting

Vote 👍  👎  
Comments for different name propositions.

### DIFF
--- a/cloud/passportauth/src/auth/PassportAuth.ts
+++ b/cloud/passportauth/src/auth/PassportAuth.ts
@@ -13,7 +13,7 @@ import { defaultDeserializeUser, defaultSerializeUser } from './UserSerializer';
  * Security interface for Raincatcher authentication middleware
  * Contains all methods that should be used to protect express routes.
  */
-export interface Auth {
+export interface EndpointSecurity {
 
   /**
    * Initializes an Express application to use passport and express-session


JameelB:
:+1:

col1985:
👍

camilamacedo86:
+1

camilamacedo86:
👍

darachcawley (changes requested):
looks good, perhaps rename the reference on line 42 too

aidenkeating (commented):
This looks fine to me

wtrocki:
**Thanks to everyone for voting**  :1st_place_medal: 
We going to incorporate changes in separate PR

aidenkeating:
@wtrocki Actually, one small comment, should it be `SecureEndpoint` rather than `EndpointSecurity`?

To me, `PassportAuth implements SecureEndpoint` reads a bit better. Could be wrong though, haven't looked into the code itself too much so it may not be doing what I am thinking.

PR #50, opened 2017-07-18
Sync data manager

wtrocki:
## Motivation

Sync itself returns data in format that is not friendly for most of the frameworks:
```javascript
{
  "uidhere": {data:"data"},
  "uidhere2": {data:"data"},
}
```

This change adds simplified sync manager that will standardize sync calls and extract uid to actual object being returned. We replicating some of the parts from old raincatcher-sync.
Managers itself will interact with any external stores we will use (like angular 2.0 observables)

When using manager resulting datasets will look like:
```javascript
[{
  uid:"uidhere", 
  data:"data"
},{
  uid:"uidhere2", 
  data:"data"
}]
```

## Progress
- [x] Sync manager
- [x] Update examples
- [x] Configure browserify build
- [x] Documentation
- [x] Unit tests
- [x] Handle data updates (needs research on angular 2.0 and observables)


client/datasync-client/example/README.md
```diff
@@ -0,0 +1,18 @@
...
+Install required packages
+
+    npm install browserify
+    npm install typescript
+    npm install tsify
+    npm install browserify-shim
+
+Build single sync bundle.
+
+    browserify --debug index.ts -p [ tsify --noImplicitAny ] > bundle.js
```
paolobueno: We could add these as devDependencies and put this as a npm script. Don't think there's any harm in having extra devDependencies for a demo.

paolobueno (dismissed):

client/datasync-client/src/config.ts
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+const SYNC_FREQUENCY = 10;
+
+export const defaultConfig = {
+  syncOptions: {
+    'syncFrequency': SYNC_FREQUENCY,
```
JameelB: this property doesn't seem to be used

paolobueno (approved):

PR #51, opened 2017-07-19
Add Keycloak Integration for Demo Cloud [WIP]

tomjackman:
## Motivation
Add Keycloak Integration for Demo Cloud

## Description
Allow Keycloak's middleware to be used in the demo cloud app to protect routes.

## Progress
- [x] Keycloak Config Definitions
- [x] Keycloak Init/Setup
- [x] Define Keycloak Config for different Enviornments


demo/server/keycloak.json
```diff
@@ -0,0 +1,9 @@
+{
+  "realm": "raincatcher",
+  "auth-server-url": "http://localhost:8080/auth",
+  "ssl-required": "external",
+  "resource": "raincatcher-cloud",
+  "public-client": true,
+  "use-resource-role-mappings": true,
```
wtrocki: Demo will use realm roles or global ones? If global roles then we do not need that.

demo/server/keycloak.json
```diff
@@ -0,0 +1,9 @@
+{
+  "realm": "raincatcher",
+  "auth-server-url": "http://localhost:8080/auth",
```
wtrocki: [Note/Optional]  best to template that and use some env variable etc.

demo/server/src/modules/index.ts
```diff
@@ -4,10 +4,13 @@ import * as express from 'express';
 import { connect as syncConnector } from './datasync/Connector';
 import { router as syncRouter } from './datasync/Router';
 import { init as authInit } from './passport-auth';
+import { init as initKeycloak } from './keycloak';
 
 export let securityMiddleware: EndpointSecurity;
 export let logger: Logger;
 
+const useKeycloak = true;
```
wtrocki: [Optional/Idea] We can connect that with existence of the environment variable that points to keycloak server.

demo/server/src/modules/index.ts
```diff
@@ -20,13 +23,27 @@ function loggerSetup() {
 }
 
 function securitySetup(app: express.Express) {
-  setupPassportSecurity(app);
+  if (useKeycloak) {
+    // user keycloak authentication
+    setupKeycloakSecurity(app);
+  } else if (process.env.NODE_ENV === 'production') {
```
wtrocki: Demo application has profile based configurations. We should not use `process.env.NODE_ENV === 'production'` directly. I'm running my node in production mode all the time because of some reasons. My vote will be to not imply this in code and just document this carefully as note.

demo/server/src/modules/keycloak/index.ts
```diff
@@ -0,0 +1,36 @@
+/// <reference path="./keycloak.d.ts" />
+
+import * as express from 'express';
+import * as session from 'express-session';
+import Keycloak = require('keycloak-connect');
+
+// Express Session Configuration.
+const sessionOpts = {
+      secret: process.env.FH_COOKIE_SECRET || 'raincatcher',
```
wtrocki: process.env.FH_COOKIE_SECRET is RHMAP specific. Let's use profile based config for this.

demo/server/keycloak.json
wtrocki: Actually we should may profile based configuration with different values for every environment lads will have.

https://github.com/feedhenry-raincatcher/raincatcher-core/blob/master/demo/server/src/util/config.ts

demo/server/src/modules/index.ts
tomjackman: This was mainly for testing purposes: I think the existence of a `keycloak.json` (with parseable JSON) in the server root folder would be a good check on whether to use Keycloak or not.

demo/server/src/modules/keycloak/index.ts
```diff
@@ -0,0 +1,36 @@
+/// <reference path="./keycloak.d.ts" />
+
+import * as express from 'express';
+import * as session from 'express-session';
+import Keycloak = require('keycloak-connect');
+
+// Express Session Configuration.
+const sessionOpts = {
```
wtrocki: [Idea] Session options seem to be common for both passport and keycloak. Maybe we can share the same options in code to make sure that we will avoid differences and possible security issues in outdated ones. ping @JameelB for opinion.

demo/server/src/modules/keycloak/index.ts
TommyJ1994: Fixed.

demo/server/src/modules/keycloak/index.ts
TommyJ1994: +1 was thinking the same. Maybe a JSON file in `/src/modules/sessionOpts.json` or similar could work.

wtrocki:
@TommyJ1994 - what's your prefference for keycloak configuration? 

Are you planning to add some automation for realm creation, 
sharing json file to be manually added or issuing docker image that is preconfigured with realm and custom login page.

I kinda think that it will be cool to have separate repository with docker image + customizations and configure image build for that. Then we can point to realm sources etc our customers to understand what is there. @JameelB @TommyJ1994 opinions on this? 
I have done that multiple times before so I can help with technical problems.

demo/server/keycloak.json
TommyJ1994: Out of the box, they will use global realms. But I was thinking if you wanted to show how a "Mobile Only" role could be created, you could show that without changing the demo apps code. That was the main reason really.

tomjackman:
@wtrocki @JameelB 

@twnolan has done work around a Keycloak Docker Image and I have created a demo Raincatcher realm for it. Repo is [here](https://github.com/twnolan/custom_keycloak_docker). It will allow you to fully configure the keycloak server for raincatcher demoing. I think thats what your looking for?

Alternatively, you can also just download the standalone server, run the start script, create an admin user, and import the realm json in the web app and you get the same result.

demo/server/src/modules/keycloak/keycloak.d.ts
```diff
@@ -0,0 +1,4 @@
+declare var KeycloakConnect:any;
+declare module "keycloak-connect" {
+    export = KeycloakConnect;
+}
```
paolobueno: :+1: Optional: we can add just a couple extra info here as the project uses it.

From what it looks like it would be something like: 
```typescript
KeycloakConnect: {
  new(options: { store: expressSession.Store }) -> KeycloakInstance
}
interface KeycloakInstance {
   middleware: (options: { logout: string }) -> express.middleware;
}
```

demo/server/src/modules/keycloak/index.ts
```diff
@@ -0,0 +1,33 @@
+/// <reference path="./keycloak.d.ts" />
+
+import * as express from 'express';
+import * as session from 'express-session';
+import Keycloak = require('keycloak-connect');
```
paolobueno: Does `import * as Keycloak from 'keycloak-connect'` work when you have the d.ts file?

I think requires are a tslint error atm, but feel free to ignore the rule with a comment if required.

demo/server/src/modules/index.ts
wtrocki: @TommyJ1994 so this means that keycloak.json we adding it's not going to be in the final demo app? Best to always have keycloak.json in the demo and just additional configuration that will point to keycloak server.

demo/server/src/modules/keycloak/index.ts
wtrocki: Good idea. Json files are hard to deal with in typescript so best to use ts/js file instead.

demo/server/src/modules/keycloak/index.ts
JameelB: @TommyJ1994 @paolobueno that works! :) We can change it to that instead of require. :+1:

demo/server/src/modules/keycloak/index.ts
TommyJ1994: Added a sessionOpts.ts file. Seems to work fine. Updated Passport to use this too.

TommyJ1994:
`Could not obtain grant code: 400:Bad Request` error fixed. 

Changed the Realm roles to Client Roles - this means that we can use `keycloak.protect('admin')` and be compliant with passport, otherwise we would need to specify the realm/client in the protect which would break the compatibility with passport.

Related PR: https://github.com/twnolan/custom_keycloak_docker/pull/6

tomjackman:
Addressed comments - moved the keycloak.json config to the `config-dev.json/config-prod.json` files. Keycloak will be used as an auth provider based on the existence of a keycloakConfig object in the `config-dev.json/config-prod.json` file, rather than a `keycloak.json` in the root folder. If the keycloakConfig doesn't exist, it will use Passport auth instead.

demo/server/src/modules/index.ts
```diff
@@ -1,13 +1,21 @@
 import { EndpointSecurity } from '@raincatcher/auth-passport';
 import { BunyanLogger, Logger } from '@raincatcher/logger';
 import * as express from 'express';
+import EnvironmentConfig, { CloudAppConfig, Config } from '../util/config';
 import { connect as syncConnector } from './datasync/Connector';
 import { router as syncRouter } from './datasync/Router';
+import { init as initKeycloak } from './keycloak';
 import { init as authInit } from './passport-auth';
 
+const appConfig: Config<CloudAppConfig> = new EnvironmentConfig<CloudAppConfig>();
```
wtrocki, line 10: In seprate PR I made this global variable so you no longer need to create new instances. 
Feel free to merge this. I will refactor it on my branch,

demo/server/src/modules/index.ts
```diff
@@ -1,13 +1,21 @@
...
+const config = appConfig.getConfig();
+
 export let securityMiddleware: EndpointSecurity;
 export let logger: Logger;
 
+// Use Keycloak if Keycloak configuration is provided
+const useKeycloak = config.keycloakConfig || false;
```
wtrocki: Value is not being used outside method. move this to `securitySetup`

demo/server/src/modules/keycloak/index.ts
```diff
@@ -0,0 +1,27 @@
+// tslint:disable-next-line:no-reference
+/// <reference path="./keycloak.d.ts" />
+
+import * as express from 'express';
+import * as session from 'express-session';
+import * as Keycloak from 'keycloak-connect';
+import EnvironmentConfig, { CloudAppConfig, Config } from '../../util/config';
+import sessionOpts from '../sessionOpts';
+
+const appConfig: Config<CloudAppConfig> = new EnvironmentConfig<CloudAppConfig>();
```
wtrocki, line 10: Same as above.

wtrocki (dismissed):

tomjackman:
Addressed comment. appConfig import changes to be addressed in another PR.

wtrocki (approved):

PR #52, opened 2017-07-19
Seed demo data

wtrocki:
## Motivation

Add ability to seed demo data. 

## Progress 

- [x] Add seeding functions
- [x] Change way we deal with id's on the server


demo/server/src/modules/datasync/Connector.ts
```diff
@@ -26,13 +28,19 @@ export function connect() {
   return new Promise(function(resolve, reject) {
     sync.connect(connectOptions, function(err, mongo, redis) {
       if (err) {
+        console.info('Error when trying to connect to mongo and redis', err);
```
paolobueno: Maybe use `console.error`?

paolobueno (dismissed):
single optional commit

package.json
```diff
@@ -27,5 +27,8 @@
     "lerna": "^2.0.0-rc.5",
     "tslint": "^5.4.3",
     "typescript": "^2.3.4"
+  },
+  "dependencies": {
+    "mongodb": "^2.2.30"
```
paolobueno: looks like you ran `npm i --save mongodb` in the root instead of the demo/server module :stuck_out_tongue:

wtrocki: That will explain why I was missing that in actual module ;)

client/logger
```diff
@@ -0,0 +1 @@
+../cloud/logger
```
paolobueno, line 1: this looks like a symlink, is it correct to include this?

demo/server/src/modules/datasync/MongoDataHandler.ts
```diff
@@ -27,8 +28,9 @@ export class GlobalMongoDataHandler {
   public setupHandleList() {
     const self = this;
     sync.globalHandleList(function(datasetId, queryParams, metadata, cb) {
+      logger.debug('GlobalHandleList called', datasetId, queryParams);
       queryParams = queryParams || {};
-      const resultPromise = this.db.collection(datasetId).find(queryParams);
+      const resultPromise = self.db.collection(datasetId).find(queryParams);
```
paolobueno, line 33: optional: can remove the need for `self` if you use arrow functions instead of closures

client/logger
wtrocki: This was attempt to make logger available for both client and cloud. 
Because we wanted to avoid duplication and another "common" category I have created symlink to that module so it will compile and will be available to consume in UI. 
Current ui has lots of console.log and we do not plan to change that but that may be useful for anyone starting to write new UI and using logger in other client side components.

paolobueno (approved):

client/logger
paolobueno: Can't we use the logger as a dependency? `@raincatcher/logger@1.0.0` inserted via lerna.

That's what `datasync-cloud` seems to be doing anyway

wtrocki: Angular repo track client packages only. Of we want this we will need to extend that to cloud which has some problems.

PR #54, opened 2017-07-21
Use development as default configuration

wtrocki:
## Motivation

Production configuration should be used only when NODE_ENV is set to production.
Otherwise we should use development.

ping @TommyJ1994


tomjackman (dismissed):
LGTM :tada:

JameelB (dismissed):

tomjackman (approved):
Working locally.

PR #56, opened 2017-07-21
Strip down wfm to bare data

paolobueno:
## Description
Strip down `@raincatcher/wfm` to bare data.
ping @wtrocki 

## Progress
- [X] Remove functional methods from the classes
- [X] Documentation
- [X] Remove outdated integration example
- [X] Postpone json schema idea
- [X] Update status names to match existing data
- [x] Update unit tests


wtrocki:
We will need to cleanup wfm integration tests

wtrocki:
@paolobueno 

Great stuff. My suggestion here is to:
- Remove/disable vehicle-inspection example as it may be confusing and it will be time consuming to adjust it.
- Adjusting tests in order for this changes to pass
- Trying to simplify this by dropping rounded statues and json schema - we can bring that in future.

wtrocki (approved):
Good work!

wtrocki:
We may create tag before merging that back. Thanks a lot!

PR #60, opened 2017-07-26
[RAINCATCH-1028] Add profile and logout route for PassportJS

JameelB:
## Motivation
The demo mobile/portal will require a `/profile` route to get user profile and a `/logout` route to log out the user. `Access-Control-Allow-Credentials` also needs to be set in the response to allow cookies to be passed in with the request for authentication. 

CORS does not allow a redirect with preflights therefore the protect() functionality of PassportJS should return a status of 401 instead.

## Description
- Add `/profile` and `/logout` route for PassportJS. 
- Set `Access-Control-Allow-Credentials` to be true for CORS requests.
- Updated Passport's protect() to send a status of 401 when the user is not authenticated

## Progress
- [x] Add `/profile` route
- [x] Add `/logout` route
- [x] Set `Access-Control-Allow-Credentials` to be true for CORS requests
- [x] Send `401` instead of a redirect on Passport's protect()
- [x] Update unit tests

## Additional Notes
Related JIRA - https://issues.jboss.org/browse/RAINCATCH-1028
Related PR - https://github.com/feedhenry-raincatcher/raincatcher-angularjs/pull/20

### DIFF
--- a/cloud/passportauth/src/auth/PassportAuth.ts
+++ b/cloud/passportauth/src/auth/PassportAuth.ts
@@ -80,9 +80,9 @@ export class PassportAuth implements EndpointSecurity {
       if (!req.isAuthenticated()) {
         if (req.session) {
           // Used for redirecting to after a successful login when option successReturnToOrRedirect is defined.
-          req.session.returnTo = req.originalUrl;
+          req.session.returnTo = req.headers.referer;
         }
-        return res.redirect(this.loginRoute);
+        return res.status(401).send();
       }
 
       const hasRole = role ? this.userService.hasResourceRole(req.user, role) : true;
--- a/cloud/passportauth/src/auth/PassportAuth.ts
+++ b/cloud/passportauth/src/auth/PassportAuth.ts
@@ -80,7 +80,7 @@ export class PassportAuth implements EndpointSecurity {
       if (!req.isAuthenticated()) {
         if (req.session) {
           // Used for redirecting to after a successful login when option successReturnToOrRedirect is defined.
-          req.session.returnTo = req.headers.referer;
+          req.session.returnTo = this.setReturnToUrl(req);
         }
         return res.status(401).send();
       }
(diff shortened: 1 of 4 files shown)


demo/server/src/app.ts
```diff
@@ -14,6 +14,16 @@ import appConfig from './util/config';
 
 const app: express.Express = express();
 const config = appConfig.getConfig();
+let corsConfig = {};
+if (!config.keycloakConfig) {
+  const dynamicOrigin = function(origin, callback) {
```
wtrocki: Looks like we changing cors options here by allowing all origins. Do we need to set cors then?

paolobueno: This is indeed equivalent to `"*"`, which is the default, so I also don't understand why we need this custom function.

We actually need to set cors for all origins because the behavior without it is not allowing cross-domain requests at all.

JameelB: @wtrocki @paolobueno Unfortunately '*' as origin is not allowed as I've set the credentials to be true in order to have the cookie in the requests. This was the reason I've put in this custom function in order to still allow cross-domain requests while allowing the credentials to be set to true.

wtrocki: This looks like critical part of the integration so let's extract that to separate method and make sure that we document what is happening here.

cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -80,7 +80,7 @@ export class PassportAuth implements EndpointSecurity {
       if (!req.isAuthenticated()) {
         if (req.session) {
           // Used for redirecting to after a successful login when option successReturnToOrRedirect is defined.
-          req.session.returnTo = req.headers.referer;
+          req.session.returnTo = this.setReturnToUrl(req);
```
wtrocki: this refers in internal function. Use self.

wtrocki (approved):

PR #61, opened 2017-07-26
Change initial data to contain proper id's

wtrocki:
## Motivation

Change initial data to contain proper id's


austincunningham (approved):
approved

PR #62, opened 2017-07-31
renamed examples to example to match all other modules

austincunningham:
## Motivation
Need the directory structure to match on all modules so scripts can be run
## Description
rename wfm/examples to wfm/example

## Progress
- [x] renamed wfm/examples to wfm/example


JameelB (approved):

wtrocki:
@austincunningham - When doing rename please do move and add to make sure that changes appear as move. This change appears as new and it's removing previous history.

PR #66, opened 2017-08-02
[RAINCATCH-1120] Update wfm entities to match data in demo apps

paolobueno:
## Motivation
For 1.0 we should keep the same nomenclature and structure for the data in RainCatcher.

## Description
- Rework proposed wfm data interfaces into 

## Progress
- [X] Reshape data interfaces
- [X] Export the new interfaces from index.ts
- [x] Move over constants related to the available step statuses
- [x] Update unit tests
- [x] Add jsdoc-based documentation to the new fields
- [x] Update README to match the entities


client/wfm/src/result/StepResult.ts
```diff
@@ -0,0 +1,17 @@
+export interface StepResult {
```
wtrocki, line 7: We forgot to migrate comments :)

paolobueno: Ah, comments for the steps?
I need to try and recreate them in the demo app.

Essentially what I did so far was grabbing data from mongo, so I'm not completely sure if I'm missing some fields or have some extra that shouldn't be on the core lib.

wtrocki: I meant actual code comments like

```
/**		
* Represents the output of a {@link Task}		
* Implementations are expected to be accompanied by a front end component to render it		
*/
```

client/wfm/src/result/StepResult.ts
```diff
@@ -0,0 +1,17 @@
...
+  submission?: {
+    [key: string]: any
+  };
+  submitter?: string;
+  timestamp: number;
+  status: string;
+  type?: string;
```
wtrocki: It's also good to document some of the fields.

client/wfm/src/result/WorkOrderResult.ts
```diff
@@ -0,0 +1,11 @@
+import {StepResult} from './StepResult';
+export interface WorkOrderResult {
+  id: string;
+  status: string;
+  nextStepIndex: number;
```
wtrocki: This field sounds like something that we actually should rename/migrate/port. WDYT?
Seen logic for `nextStepIndex` and I do not think that will be a problem:
https://github.com/feedhenry-raincatcher/raincatcher-angularjs/search?utf8=%E2%9C%93&q=nextStepIndex&type=

client/wfm/src/result/WorkOrderResult.ts
```diff
@@ -0,0 +1,11 @@
+import {StepResult} from './StepResult';
+export interface WorkOrderResult {
```
wtrocki: In the original format this was named without middle capital letter = `Workorder..` but I'm happy to have it this way for TypeScript stuff.

client/wfm/src/step/Step.ts
```diff
@@ -2,13 +2,17 @@
...
    * Code that matches the identifier in {@link TaskHandler}
    */
   code: string;
+  name: string;
   /**
-   * Options object that act as runtime parameters for its equivalent {@link Task}
+   * Options object that act as runtime parameters for its equivalent {@link Step}
    */
-  options?: object;
+  templates?: {
```
wtrocki: +1 for having this still in the object for the moment until we figure out how to migrate this to templates being part of the step. Checking code usages suggest that this may be a non trivial task:
https://github.com/feedhenry-raincatcher/raincatcher-angularjs/search?utf8=%E2%9C%93&q=templates.form&type=

client/wfm/src/result/WorkOrderResult.ts
wtrocki: Add documentation

client/wfm/src/workflow/WorkFlow.ts
```diff
@@ -0,0 +1,17 @@
+import {Step} from '../step/Step';
+import {WorkOrder} from '../workorder/WorkOrder';
+
+/**
+ * Definition holder for a linear set of {@link Task}s
```
wtrocki, line 5: Outdated docs

client/wfm/src/workorder/WorkOrder.ts
```diff
@@ -1,14 +1,19 @@
 import * as Promise from 'bluebird';
 import { min } from 'lodash';
-import { getRoundedStatus, Task, TaskStatus } from '../task/Task';
+
+type LatLng = [number, number];
+
 /**
  * The executable instance of a {@link Process}
```
wtrocki: Outdated docs.

client/wfm/src/workorder/WorkOrder.ts
```diff
@@ -1,14 +1,19 @@
...
  */
-export interface ProcessInstance {
+export interface WorkOrder {
   /**
    * Unique identifier for this ProcessInstance
    */
   id: string;
+
+  type: string;
```
wtrocki: Let's document this. Cannot find for what we use this type.  It looks like it's used only for label so `label` or name will be more appropriate:
https://github.com/feedhenry-raincatcher/raincatcher-angularjs/search?utf8=%E2%9C%93&q=workorder.type&type=

client/wfm/src/workorder/WorkOrder.ts
```diff
@@ -28,17 +33,16 @@ export interface ProcessInstance {
   /**
    * Longer description for this {@link ProcessInstance}
    */
-  comment: string;
+  summary: string;
```
wtrocki: Actually that's the data we do not want here as having this will imply that user workorders will have location etc. Can we use some generic data field like `attributes` that users can use to post any content. We can adjust UI to reflect to this and this data will be actually demo specific. 

💯  I'm really against having this in framework.

wtrocki (dismissed):

client/wfm/src/result/StepResult.ts
paolobueno: ah, I hadn't updated those yet because this was still WIP, so I thought you were talking about something else :+1:

client/wfm/src/result/WorkOrderResult.ts
paolobueno: I think the solution should only move to the next step by default, and have specialized steps for branching.

Right now I think the most we can do with this field is move it to another entity like `workorder.nextStepIndex`, though I'm not sure it's worth it.
Bigger changes in this logic should be in another ticket I guess.

client/wfm/src/workorder/WorkOrder.ts
paolobueno: Yes I think this one is just written-to never read. I'll just remove it for now

client/wfm/README.md
```diff
@@ -1,24 +1,18 @@
 ## RainCatcher WFM
 
-RainCatcher **W**ork**f**low **M**anagement module is an TypeScript implementation for a subset of the [BPMN specification](http://www.bpmn.org/).
+RainCatcher **W**ork**F**low **M**anagement module is an TypeScript implementation for a equivalent subset of the [BPMN specification](http://www.bpmn.org/).
```
JameelB, line 3: is `a` TypeScript implementation for `an` equivalent

client/wfm/README.md
```diff
@@ -1,24 +1,18 @@
...
-* `Executor`: transforms Process into ProcessInstance and allows to control transitions between `Tasks`
-
-Users can write custom `Task`s that will implement any business logic and link implementation with any UI solution.
+* `WorkFlow`: contain an ordered set of definitions of `Step`s, to be instantiated into a `WorkOrder` for execution
+* `WorkOrder`: instance of a `WorkFlow` to be executed
+* `Step`: defines single unit of work
+* `StepResult`: the output of the execution of a `Step`
+* `WorkOrderResult`: The tracking of the progress of a `WorkOrder`
 
 THe classes are framework-agnostic, allowing developers to focus on the implementation of Business Process on top of any new or existing solution.
```
JameelB, line 17: capital H on `THe`

client/wfm/src/status.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * List of available status for available for {@link WorkOrder}s and their individual {@link Step}s
```
JameelB, line 2: remove `for` in `available for {@link WorkOrder}s`

client/wfm/src/step/Step.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * A single unit of work, containing its identifier and runtime options for customizing behavior.
+ */
+export interface Step {
+  /**
+   * Code that matches the identifier of the module that should handle the execution of this {@link Step}
+   */
+  code: string;
+  name: string;
```
JameelB, line 9: Missing name description?

paolobueno:
closing in favor of new pr along with changes in #71

PR #71, opened 2017-08-04
[RAINCATCH 1121] migrate wfm service to core

paolobueno:
## Motivation
Have the wfmService on core since it's a central part of the app logic, set a base for documentation on top of TS for it.

## Progress
- [X] Extract wfmService from angularjs repo
- [X] Add type annotations
- [X] Include interface info for data services
- [x] jsdocs
- [ ] Unit tests
- [x] Documentation


paolobueno:
ping @wtrocki

paolobueno:
ping @wtrocki

client/wfm/src/service/DataService.ts
```diff
@@ -0,0 +1,8 @@
+import * as Promise from 'bluebird';
+export interface DataService<T> {
```
wtrocki, line 2: This looks like candidate for the data storage client module :)

client/wfm/src/service/WfmService.ts
```diff
@@ -0,0 +1,241 @@
...
+import { WorkOrderResult } from '../result/WorkOrderResult';
+import { STATUS } from '../status';
+import { Step } from '../step/Step';
+import { WorkFlow } from '../workflow/WorkFlow';
+import { WorkOrder } from '../workorder/WorkOrder';
+import { DataService } from './DataService';
+import { ResultService } from './ResultService';
+import { UserService } from './UserService';
+
+interface CompleteStepParams {
```
wtrocki: IMHO name is strange. `CompleteStepParams` => `CompleteStepData`

client/wfm/src/service/WfmService.ts
```diff
@@ -0,0 +1,241 @@
...
+  /**
+   * Beginning a workflow for a single Workorder.
+   *
+   * @param {string} workorderId - The ID of the workorder to begin the workflow for.
+   */
+  public beginWorkflow(workorderId) {
+    return this.workorderSummary(workorderId).then(summary => {
+      const workorder = summary.workorder;
+      const workflow = summary.workflow;
+      const res = summary.result || this.createNewResult(workorderId, workorder.assignee);
```
wtrocki: It's not clear when result will be already available. It looks like hack for some bug and we should always create new result here.

client/wfm/src/service/WfmService.ts
```diff
@@ -0,0 +1,241 @@
...
+
+      // When the result has been read/created, then we can move on.
+      return Promise.resolve(res).then(result => {
```
wtrocki: We were saying that framework code should be really callback based and we can wrap things into promises in clients. This will bring some inconsistence, comparing to the other modules we have at the moment. 

Do you think we should made exception for client side modules ? I'm really open for this, but flagging it as possible issue.

client/wfm/src/service/WfmService.ts
```diff
@@ -0,0 +1,241 @@
...
+        workorderResult.status = this.checkStatus(workorder, workflow, workorderResult);
+        workorderResult.nextStepIndex = this.executeStepReview(workflow.steps, workorderResult).nextStepIndex;
+
+        return this.resultService.update(workorderResult).then(function() {
+          return {
+            workorder,
+            workflow,
+            result: workorderResult,
+            nextStepIndex: workorderResult.nextStepIndex,
+            step: workorderResult.nextStepIndex > -1 ? workflow.steps[workorderResult.nextStepIndex] : workflow.steps[0]
```
wtrocki: That looks terrible and may cause some issues (saw things crashing in this area)

client/wfm/src/service/WfmService.ts
```diff
@@ -0,0 +1,241 @@
...
+          status: STATUS.COMPLETE,
+          timestamp: new Date().getTime(),
+          submitter: profileData.id
+        };
+
+        // The result needs to be updated with the latest step results
+        workorderResult.stepResults = workorderResult.stepResults || {};
+        workorderResult.stepResults[step.code] = stepResult;
+        workorderResult.status = this.checkStatus(workorder, workflow, workorderResult);
+        workorderResult.nextStepIndex = this.executeStepReview(workflow.steps, workorderResult).nextStepIndex;
```
wtrocki: reformat this code. It's strange that we relay on return value this way.

client/wfm/src/service/WfmService.ts
```diff
@@ -0,0 +1,241 @@
...
+    });
+  }
+
+  /**
+   * Checks each of the result steps to determine if the workflow is complete,
+   * and if not, what is the next step in the workflow to display to the user.
+   */
+  protected executeStepReview(steps: Step[], result: WorkOrderResult) {
+    // tslint:disable-next-line:max-line-length
+    // See https://github.com/feedhenry-raincatcher/raincatcher-workflow/blob/b515e8acefad4bc50a7cc281863e2176c8babbed/lib/client/workflow-client/workflowClient.js
```
wtrocki: Remove this. Let's not point to some old code.

client/wfm/src/service/WfmService.ts
```diff
@@ -0,0 +1,241 @@
...
+   * Checking the status of a workorder
+   *
+   * @param workorder  - The workorder to check status
+   * @param workflow   - The workflow to check status
+   * @param result     - The result to check status
+   */
+  protected checkStatus(workorder: WorkOrder, workflow: WorkFlow, result: WorkOrderResult): STATUS {
+    let status;
+    const stepReview = this.executeStepReview(workflow.steps, result);
+    if (stepReview.nextStepIndex >= workflow.steps.length - 1 && stepReview.complete) {
```
wtrocki, line 237: That doesn't sound like object oriented aproach. Maybe we can refactor this into separate class. StatusManager and allow people to override this if different statuses are done. I will want to do this also for `nextStepIndex` logic. It looks like nextStepIndex is used only for this specific logic, I believe we can also extract it.

To not overload you with this it can be done as separate PR, but I think we need that for our extensibility story.

client/wfm/src/service/DataService.ts
paolobueno: Yes, this would be the interface the user needs to implement to have another, non-sync-based storage.

Just the result service right now needs an extra method, and that's reflected in the constructor for the wfm service.

client/wfm/src/service/WfmService.ts
paolobueno: They're the parameters for the function: https://github.com/feedhenry-raincatcher/raincatcher-core/blob/RAINCATCH-1121-corewfm/client/wfm/src/service/WfmService.ts#L127, not a return value, only here for documentation.

Alternatively we can change the function to receive the 3 params separately instead of through an object.

paolobueno:
closing in favor of new pr along with changes in #66

paolobueno:
done by #74

PR #74, opened 2017-08-08
[RAINCATCH 1120+1121] Core wfm integration

paolobueno:
## Motivation

Have the wfmService on core since it's a central part of the app logic, set a base for documentation on top of TS for it.
Update the entities in core/wfm to match data used by the demo applications.

## Description
Merger of #66 and #71

## Progress
- [X] Export the new interfaces from index.ts
- [x] Move over constants related to the available step statuses
- [x] Update unit tests
- [x] Add jsdoc-based documentation to the new fields
- [x] Update README to match the entities
- [X] Extract wfmService from angularjs repo
- [X] Add type annotations
- [X] Include interface info for data services
- [X] jsdocs
- [X] Reshape data interfaces
- [x] Unit tests

## Additional Notes

Companion PR is https://github.com/feedhenry-raincatcher/raincatcher-angularjs/pull/32


client/wfm/src/result/StepResult.ts
```diff
@@ -0,0 +1,35 @@
...
+import { Step } from '../step/Step';
+
+/**
+ * Output of a single {@link Step} in the context of a {@link WorkOrder}
+ */
+export interface StepResult {
+  /**
+   * Form data output of the {@link Step}'s execution
+   */
+  submission?: {
```
wtrocki, line 11: This format is really strange and was added to reference submissions by type. In practice this may lead to multiple issues (for example when same key is used twice). I think we may pull out key into object to be more object friendly:
```typescript
{
key:string, 
data:any
}
```

client/wfm/src/result/StepResult.ts
```diff
@@ -0,0 +1,35 @@
...
+  };
+  /**
+   * Identifier of the User that submitted the form data
+   */
+  submitter?: string;
+
+  /**
+   * Unix timestamp of the creation of this result
+   */
+  timestamp: number;
```
wtrocki, line 22: We may document how we store date in system later (timestamp) aproach and how to work with that in UI (using moment etc.)

client/wfm/src/result/StepResult.ts
wtrocki: We may even create submission interface to help people to create Steps and denote what should be send as submissions - extra fields etc.

client/wfm/src/service/DataService.ts
```diff
@@ -0,0 +1,8 @@
+import * as Promise from 'bluebird';
+export interface DataService<T> {
```
wtrocki: Documentation needed. Really important for this.

client/wfm/src/service/DataService.ts
```diff
@@ -0,0 +1,8 @@
...
+  list(): Promise<T[]>;
+  read(id): Promise<T>;
+  create(item: T): Promise<T>;
+  update(item: T): Promise<T>;
+  remove(item: T): Promise<T>;
```
wtrocki: Let's have agreement on callback vs promises for client facing api.

client/wfm/src/service/WfmService.ts
```diff
@@ -0,0 +1,248 @@
...
+/**
+ * Return value of the {@link WfmService#workorderSummary} interface
+ */
+export interface Summary {
+  workflow: WorkFlow;
+  workorder: WorkOrder;
+  result: WorkOrderResult;
+  /**
+   * Index of the next {@link Step} the {@link WorkOrder} should progress to after completion of the current one.
+   * This allows for skipping or returning to a previous one.
```
wtrocki: `This allows for skipping or returning to a previous one.` => previous one => `previous step`

client/wfm/src/service/WfmService.ts
```diff
@@ -0,0 +1,248 @@
...
+   */
+  nextStepIndex: number;
```
wtrocki: Why this cannot be part of the workorder? I see that being used only in 2 places so this change will be trivial and make more cleaner api. `

client/wfm/src/service/WfmService.ts
```diff
@@ -0,0 +1,248 @@
...
+          // Now we check the current status of the workflow to see where the next step should be.
+          result.status = this.checkStatus(workorder, workflow, result);
+
+          // We now have the current status of the workflow for this workorder, the begin step is now complete.
+          return {
+            workorder,
+            workflow,
+            result,
+            nextStepIndex,
+            step: nextStepIndex > -1 ? workflow.steps[nextStepIndex] : workflow.steps[0]
```
wtrocki: Extract this to method or improve - it's really hard to read and extract what is happening here.

client/wfm/src/result/WorkOrderResult.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Identifier of the User assigned to the {@link WorkOrder}
+   */
+  assignee: string;
+
+  /**
+   * Set of {@link StepResult}s for the {@link Step}s contained in the related {@link WorkOrder}
+   */
+  stepResults?: {
```
wtrocki, line 31: Same comment as for submissions. I think format is terrible and it was just done as shortcut to avoid controller specific method (this is referenced in UI)

client/wfm/src/service/WfmService.ts
```diff
@@ -0,0 +1,248 @@
...
+      ]).then(([workflow, result]) => ({
+        workflow,
+        workorder,
+        result,
+        nextStepIndex: this.executeStepReview(workflow.steps, result).nextStepIndex
+      })));
+  }
+
+  /**
+   *
```
wtrocki: --\n :)

client/wfm/src/service/WfmService.ts
```diff
@@ -0,0 +1,248 @@
...
+      let { nextStepIndex } = summary;
+
+      if (!result) {
+        // No result exists, The workflow should have been started
+        return Promise.reject(new Error('No result exists for workflow ' + workorderId +
+          '. The workflow back topic can only be used for a workflow that has begun'));
+      }
+
+      // -1 is a special value for 'no next step'
+      nextStepIndex = _.max([nextStepIndex - 1, -1]) || -1; // _.max returns `number?`
```
wtrocki, line 141: I think that nextStepIndex here is crucial point, that may be problematic to do things we plan in future. 
If we code it so it will be state machine. I would really like nextStepIndex to disappear and apply rule decision facade that will change current step. Doing this math operations doesn't contribute to visiblity and flexibility.
Imagine that someone want to add some conditions to decide where to go. 

I think I have idea how to make it flexible and clean. Let's agree on something and possibly create new ticket. I will be able to migrate from `nextStepIndex` fairly fast by adding some simple `StepTransitionEngine` that can be extended in future (for example to meet BPMn spec)

client/wfm/src/service/WfmService.ts
```diff
@@ -0,0 +1,248 @@
...
+
+  /**
+   * Completing a single step for a workorder.
+   */
+  public completeStep(parameters: CompleteStepParams): Promise<StepOperationData> {
+    const workorderId = parameters.workorderId;
+    const stepCode = parameters.stepCode;
+    const submission = parameters.submission;
+    return Promise.all([
+      this.userService.readUser().then(profileData => profileData.id),
```
wtrocki, line 162: @JameelB FYI - we using userService in the WFM module

client/wfm/src/service/WfmService.ts
```diff
@@ -0,0 +1,248 @@
...
+          ' and step code ' + stepCode));
+      }
+
+      // Got the workflow, now we can create the step result.
+      const stepResult = {
+        step,
+        submission,
+        status: STATUS.COMPLETE,
+        timestamp: new Date().getTime(),
+        submitter: userId
```
wtrocki, line 188: `submitter` imho is sounds strange. If that's not a big task we may change it to `completedBy`, `owner`, `userReference` etc.

client/wfm/README.md
```diff
@@ -29,19 +23,3 @@ In order to use this module, begin by installing all dependencies:
 If inside the [lerna](http://lernajs.io)-based mono repo: `lerna bootstrap --scope=@raincatcher/wfm --include-filtered-dependencies`
 
 For standalone usage: `npm install`
-
-For implementing a base flow, see the [examples folder](./examples/).
```
wtrocki, line 33: We missing example code (I know it may be strange to do this but as we agreed on we need to have examples and try to document some specifics like new step creation etc.)

client/wfm/src/result/StepResult.ts
paolobueno: Not sure what you mean by this.
```
submission: {
    "fuel" : 25,
    "tires" : true,
    "lights" : true
}
```
Is sample data from vehicle inspection, currently the form model being passed from angular.
What do you think it should shape like? Repeated keys should be invalid I guess.

I think this should be in a new issue.

client/wfm/src/service/WfmService.ts
paolobueno: This was in the result data, now it's only calculated. Do you think it should be stored?

client/wfm/src/service/DataService.ts
wtrocki: Offline conversation - we decided to leave WFM with promissified api.

client/wfm/src/service/WfmService.ts
wtrocki: Let's create new ticket to get rid of this stepIndex to allow building of some flexible flows.

client/wfm/README.md
paolobueno: I don't think we have step creation set in stone yet, I just removed the outdated decription that I had for the previous version with the names like `Process/ProcessInstance` etc, because their usage was to be different.

wtrocki (approved):

PR #75, opened 2017-08-08
WFM api framework - HTTP RESTfull api for WFM (ex-webapi)

wtrocki:
## Motivation

Allow to use express based API to operate on WFM objects.

Note: this changes are WFM specific and it will require integration with WFM module = sparate PR in angular.js repository

## Notes

I have done some investigation for pagination and I did not found anything that will meet all of our criteria. I wrote simple pagination mechanism for this module. We can extract this to separate module if needed somewhere else but I think it makes sense to have it here. 

## Progress
- [x] Api definitions
- [x] Extract to separate module
- [x] Pagination
- [x] Security
- [x] Documentation
- [x] Unit tests


demo/server/src/modules/wfm-web-api/DataService.ts
```diff
@@ -0,0 +1,35 @@
...
+  /**
+   * @param db - mongodb driver connection
+   * @param collectionName - name of the collection stored in mongodb
+   */
+  constructor(readonly db: Db, readonly collectionName: string) {
+  }
+
+  public list() {
+    return this.db.collection(this.collectionName).find({});
+  }
```
paolobueno: I think this class, or an interface definition of it should be async, not all storage libs can fit a synchronous api.

demo/server/Gruntfile.js
```diff
@@ -14,6 +14,7 @@ module.exports = function (grunt) {
         },
         exec: {
             'run': 'ts-node src/index.ts',
+            'runDebug': 'ts-node --inspect src/index.ts',
```
wtrocki, line 17: @austincunningham We can debug now.

cloud/wfm-rest-api/example/index.ts
```diff
@@ -0,0 +1,33 @@
...
+app.use('/api', api.createWFMRouter());
+
+// Use connect method to connect to the server
+const url = 'mongodb://localhost:27017/raincatcher';
+MongoClient.connect(url, function(err, db) {
+  if (db) {
+    api.setDb(db);
+  } else {
+    console.info('MongoDb not connected', err);
+    process.exit();
```
paolobueno: Should add some non-zero value to make sure the process exits with an error: `process.exit(1)`.

Alternatively this could be done before starting the http server at all:
`promisify(MongoClient.connect)(...).then(db => api.setDb(db)).then(() => app.listen(...));`

cloud/wfm-rest-api/src/ApiConfig.ts
```diff
@@ -0,0 +1,25 @@
...
+  resultApiName?: string;
+  workorderCollectionName?: string;
+  workflowCollectionName?: string;
+  resultCollectionName?: string;
+}
+
+/**
+ * Default module configuration
+ */
+export const defaultConfiguration = {
```
paolobueno: could `implement ApiConfig`, just to inherit jsdoc info

cloud/wfm-rest-api/src/ApiConfig.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Module configuration interface
+ * Holds all values that can be changed to modify module behavior
+ */
+export interface ApiConfig {
+  workorderApiName?: string;
+  workflowApiName?: string;
+  resultApiName?: string;
+  workorderCollectionName?: string;
+  workflowCollectionName?: string;
+  resultCollectionName?: string;
```
paolobueno: need jsdocs for these keys

cloud/wfm-rest-api/src/WfmRestApi.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  /**
+   * Create new router for hosting WFM http api.
+   */
+  public createWFMRouter() {
+    const router: express.Router = express.Router();
+    const workorderController = new ApiController(router, this.workorderService, this.config.workorderApiName);
+    const workflowController = new ApiController(router, this.workflowService, this.config.workflowApiName);
+    const resultController = new ApiController(router, this.resultService, this.config.resultApiName);
```
paolobueno: -1 for extra class here, this is using a constructor only for side-effects (mutating `router`).

Just need something like 
```typescript
router.use(`/${workorderApiName}`, buildRouter(this.workorderService))
```
so you don't need to mutate router inside another function, and reduces # of params for buildRouter.

cloud/wfm-rest-api/src/data-api/CrudRepository.ts
```diff
@@ -0,0 +1,51 @@
...
+import { PageResponse } from '../data-api/PageResponse';
+
+/**
+ * Interface for building CRUD database repository.
+ * Clients should implement this interface to connect storage of WFM objects
+ * with different data storage engines like MySQL etc.
+ * Interface is being used internally to perform database operations for WFM models
+ * It's not recomended to use it for other application business logic.
+ */
+export interface CrudRepository {
```
paolobueno: I don't like using the term CRUD itself, feels like too much slang and signals boilerplate code.

Could be `PagingDataRepository` or something, much fancier :D

cloud/wfm-rest-api/src/data-api/PaginationEngine.ts
```diff
@@ -0,0 +1,88 @@
...
+import { PageResponse } from '../data-api/PageResponse';
+
+/**
+ * Pagination procesor
+ * Clients may override this class to provide custom pagination parameters
+ *
+ * Note: pages are counted from number 1.
+ * Page 0 requests will return empty data.
+ */
+export class PaginationEngine {
```
paolobueno: Maybe `MongoPaginationEngine`? Implementation is mongo-specific.

cloud/wfm-rest-api/src/impl/ApiController.ts
```diff
@@ -0,0 +1,120 @@
...
+      }
+
+      repository.delete(req.params.id).then(function(data) {
+        res.send(data);
+      }).catch(function(err) {
+        self.errorHandler(req, res, err);
+      });
+    });
+  }
+  protected errorHandler(req: express.Request, res: express.Response, error: ApiError) {
```
paolobueno: We don't need a local error handler for this router.
This would prevent the application to define it's own global error handler, and even try to handle errors from this module in a different way.

We do have a global error handler in the app by the way: https://github.com/feedhenry-raincatcher/raincatcher-core/blob/master/demo/server/src/app.ts#L66

paolobueno: For example, production error handlers can redirect a user to a 'oops' page, log exceptions to external storage or email, etc.

demo/server/Gruntfile.js
paolobueno: Why do we need grunt over this same line in an npm script?

paolobueno: Ah, I guess it's for reloading the app? There's `nodemon` for that too

demo/server/src/modules/keycloak/index.ts
```diff
@@ -1,8 +1,8 @@
 import * as express from 'express';
 import * as session from 'express-session';
 
-import appConfig from '../../util/config';
-import sessionOpts from '../sessionOpts';
+import appConfig from '../../util/Config';
```
paolobueno: I still see this file as lowercase, dunno how this works for you!

cloud/wfm-rest-api/src/WfmRestApi.ts
wtrocki: 👍 That will be also easy to unit tests. Do you see api exposing each crud method or we should get entire CRUD operations as now?

cloud/wfm-rest-api/src/impl/ApiController.ts
wtrocki: Not sure what you suggesting here. I wanted this module to be responsible for error handling as app currently tries to render html.  This will throw nice json.

What will be your suggestion - throw error here ?

demo/server/src/modules/keycloak/index.ts
wtrocki: I have it uppercase but git somehow did not pushed that change. Sorry

cloud/wfm-rest-api/src/impl/ApiController.ts
paolobueno: Ah, in order to allow the top level app to use the global error handler you add the third param and call `next(err)`, the app might crash on uncaught errors.

The global error handler that I linked should return a formatted json just the same, but it's possible to override it at that global point.

See http://expressjs.com/en/guide/error-handling.html, more specifically:
```javascript
    PaidContent.find(function (err, doc) {
      if (err) return next(err)
      res.json(doc)
    })
```

cloud/wfm-rest-api/src/WfmRestApi.ts
paolobueno: I think it's okay to return a router with all the methods, the idea behind `router.use('/mount-point', middleware)` is that that middleware can properly handle everything from that point onwards.

demo/server/src/modules/keycloak/index.ts
wtrocki: Done

cloud/wfm-rest-api/src/impl/ApiController.ts
wtrocki: I just do not wanted this module to be self sufficient and do not leak some of the elements to users top level application, but this aproach will be more aligned with express patterns.

demo/server/config-dev.json
```diff
@@ -1,7 +1,10 @@
 {
   "morganOptions": null,
   "logStackTraces": true,
-  "sync":{
+  "security": {
+    "apiRole": "ADMIN"
```
wtrocki: @JameelB is that `admin` or `ADMIN` ?

JameelB: it's `admin`

cloud/wfm-rest-api/package.json
```diff
@@ -29,8 +29,8 @@
     "report-dir": "coverage_report",
     "check-coverage": true,
     "lines": 75,
-    "functions": 100,
-    "branches": 80
+    "functions": 80,
```
wtrocki: 100 it's just almost impossible

demo/server/Gruntfile.js
wtrocki: hmm.. actually it's for life reload that was supported in previous application..
I couldn't really get that running so life reload will not work now :(

cloud/wfm-rest-api/README.md
```diff
@@ -0,0 +1,108 @@
+# RainCatcher Api module
+
+Module used to expose express based api for WFM objects.
+
+### WFM speficic implementations
```
JameelB: `specific`

cloud/wfm-rest-api/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Custom database integrations
+
+Custom database integrations are possible thanks to `CrudRepository` interface.
+See Api documentation for more details.
+
+## Rest API
+
+Module provides a way to dynamically create API for different business objects.
+Created api will use simplied implementations for typical create, read, delete and update operations.
```
JameelB: `simplified`

cloud/wfm-rest-api/README.md
```diff
@@ -0,0 +1,108 @@
...
+It's not recomended to use it outside WFM framework to store user related objects. Please use mongodb driver directly.
+
+## Rest API definitions
+
+Definitions apply to every to object exposed by this API. Placeholder `{object}` can be replaced by `workflow`, `workorder` and `result`.
```
JameelB: remove `to` in `every to object`

cloud/wfm-rest-api/example/README.md
```diff
@@ -0,0 +1,15 @@
+## wfm-web-api example
+
+### Requirements
+
+Mongodb running on standard port
+
+### Runining example
```
JameelB: `Running`

demo/server/Gruntfile.js
austincunningham: :)

cloud/wfm-rest-api/src/data-api/ApiError.ts
```diff
@@ -0,0 +1,9 @@
+
+/**
+ * Interface used to construct standarized response for api error handlers.
```
JameelB: `standardized`

JameelB: `standardized`

cloud/wfm-rest-api/src/impl/ApiController.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * Delete handler
+   * Can be reused by developers that wish to mount handler directly on router
+   */
+  public deleteHandler(req: Request) {
+    getLogger().debug('Api delete method called', { params: req.params });
+
+    if (!req.params.id) {
+      const error = new ApiError(errorCodes.MISSING_ID, 'Missing id parameter');
```
JameelB: Missing statusCode `400`?

cloud/wfm-rest-api/src/impl/ApiController.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * Update handler
+   * Can be reused by developers that wish to mount handler directly on router
+   */
+  public putHandler(req: Request) {
+    getLogger().debug('Api update method called', { body: req.body });
+
+    if (!req.body) {
+      const error = new ApiError(errorCodes.CLIENT_ERROR, 'Missing request body');
```
JameelB: Missing statusCode `400`?

cloud/wfm-rest-api/src/data-api/ApiError.ts
```diff
@@ -0,0 +1,9 @@
...
+ */
+export class ApiError extends Error {
+  constructor(public code: string, public message: string, statusCode: number = 500) {
```
JameelB: code and statusCode not used?

wtrocki: TypeScript magic. Adding public is exposing them as fields.

JameelB: may need to add `public` to statusCode as well?

cloud/wfm-rest-api/README.md
jstaffor: Possibly better to say
"create, read, update and delete" - think of CRUD

cloud/wfm-rest-api/README.md
```diff
@@ -0,0 +1,108 @@
...
+### Custom database integrations
+
+Custom database integrations are possible thanks to `CrudRepository` interface.
+See Api documentation for more details.
+
+## Rest API
+
+Module provides a way to dynamically create API for different business objects.
+Created api will use simplied implementations for typical create, read, delete and update operations.
+It's not recomended to use it outside WFM framework to store user related objects. Please use mongodb driver directly.
```
jstaffor: 1) What does "it" refer to in "to use it" - the rest api?
2) outside "the" WFM framework

cloud/wfm-rest-api/README.md
```diff
@@ -0,0 +1,108 @@
...
+Additionaly error metadata is being returned:
+
+```json
+{
+  "code":"InvalidID",
+  "message":"Provided id is invalid"
+}
+```
+
+> **Note:** If you apply security middleware additional `401` and `403` statuses may be returned
```
jstaffor: Suggestion: "If you apply middleware security, additional "

cloud/wfm-rest-api/example/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+    ts-node example/index.ts
+
+To test example try
+
+    curl http://localhost:3000/api/workorders
+
+Note: This example assumes that you have run demo application that populated your database.
```
jstaffor: ...run "the" demo application...

cloud/wfm-rest-api/src/ApiConfig.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Module configuration interface
+ * Holds all values that can be changed to modify module behavior
+ */
+export interface ApiConfig {
+  /** Used to create workorder route */
+  workorderApiName?: string;
+  /** Used to create workflow route */
+  workflowApiName?: string;
+  /** Used to create result route */
+  resultApiName?: string;
```
paolobueno: I think we can get rid of the `Used to`s here, and more detail for the client developer: `Root path for the Result http endpoints.`

austincunningham (approved):
verified

paolobueno (approved):

PR #82, opened 2017-08-14
Adding cloud auth for sync

wtrocki:
## Motivation
 Adding cloud auth for sync

## Progress
- [x] Protect sync calls

### DIFF
--- a/demo/server/config-dev.json
+++ b/demo/server/config-dev.json
@@ -2,7 +2,8 @@
   "morganOptions": null,
   "logStackTraces": true,
   "security": {
-    "apiRole": "admin"
+    "apiRole": "admin",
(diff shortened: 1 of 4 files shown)


wtrocki:
@JameelB we may need to add here the same pattern for role as in https://github.com/feedhenry-raincatcher/raincatcher-core/pull/75

paolobueno (approved):

wtrocki:
Most of the changes here were made by @JameelB.
It looks like protect works only for passport. There is generic problem with keycloak that will be resolved in separate issue. I have created separate ticket to cover this work.

wtrocki (commented):
Reviewed, Verified.

PR #84, opened 2017-08-14
hotfix update route, add morgan and error logging line

paolobueno:
## Motivation
Client expected to send update requests to `PUT /{entity}/:id` while server code only supported it on the root of the entity path, we can support both uses.

## Description
- Add support for `PUT /:id`
- Add some logging


cloud/wfm-rest-api/src/impl/ApiController.ts
```diff
@@ -95,6 +95,12 @@ export class ApiController<T> {
       return Bluebird.reject(error);
     }
 
+    req.body.id = req.body.id || req.params.id;
+    if (!req.body) {
```
wtrocki: Do you mean `if (!req.body.id) {`?

cloud/wfm-rest-api/src/impl/ApiController.ts
```diff
@@ -95,6 +95,12 @@ export class ApiController<T> {
       return Bluebird.reject(error);
     }
 
+    req.body.id = req.body.id || req.params.id;
```
wtrocki: I'm not sure we should be flexible here. Which way do we prefer? 
Documentation states that api for put is `PUT /`. If we want to expose id as param let's update this.

cloud/wfm-rest-api/src/impl/ApiController.ts
```diff
@@ -121,7 +127,8 @@ export class ApiController<T> {
       .put(this.buildExpressHandler(this.putHandler));
     router.route('/:id')
       .get(this.buildExpressHandler(this.getHandler))
-      .delete(this.buildExpressHandler(this.deleteHandler));
+      .delete(this.buildExpressHandler(this.deleteHandler))
+      .put(this.buildExpressHandler(this.putHandler));
```
wtrocki: `put` is added twice. That's really misleading.

cloud/wfm-rest-api/src/impl/MongoDbRepository.ts
```diff
@@ -13,7 +14,10 @@ const dbError: ApiError = new ApiError(errorCodes.DB_ERROR, 'MongoDbRepository d
 /**
  * Service for performing data operations on mongodb database
  */
-export class MongoDbRepository<T extends { id: string }> implements PagingDataRepository<T> {
+export class MongoDbRepository<T extends {
+  id: string,
+  _id?: string | ObjectID
```
wtrocki, line 19: [Question] Why we need that? It looks strange.

wtrocki (changes requested):
Create standardized api for `PUT`.  Adjust documentation. 
Let's avoid overcomplicated generics.

cloud/wfm-rest-api/src/impl/ApiController.ts
paolobueno: Let's have `PUT /:id`, it's the regular way to go since it's a request to modify the resource at the endpoint to a specific entity.

see section 9.6 [here](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html)

cloud/wfm-rest-api/src/impl/ApiController.ts
paolobueno: one for the root, one for `/:id`, I'll remove the first one

cloud/wfm-rest-api/src/impl/MongoDbRepository.ts
paolobueno: When this repository is grabbing the data from mongo it comes with `_id: ObjectID('...')`, but it's serialized to a string when we try to save it again and mongo complains that it won't change it.

We can either `delete data._id` like I did or try to wrap it again in an `ObjectID` constructor and try to update it back to the same value. I had some trouble trying out the second option.

wtrocki: I think we have two options. Move all queries to use _id. or remove _id completely from every query. Second option seems better to me and align with the was sync work.
I think I also used wrong syntax for update https://stackoverflow.com/questions/22356110/update-field-inside-document-by-id-in-mongodb

Another thing will be to grab that obj into interface. This looks really strange .

wtrocki (dismissed):

wtrocki (dismissed):

wtrocki (approved):

PR #85, opened 2017-08-17
[RAINCATCH-1164] fix redirect bug after login when already authenticated

JameelB:
## Motivation
When logging into a demo app when you're already logged into another, it will redirect you to http://localhost:8001/ which will render `{ name: "raincatcher", version: "1.0.0"}`. We need to check if the user is already authenticated before calling Passport's authenticate to redirect back to the application.

## Description
Once a user has already successfully logged in, Passport deletes *req.session.returnTo* field which causes Passport to redirect to the default URL which is currently '/'. Check if the user is already authenticated before calling Passport's authenticate function. 

## Progress
- [x] Add isAuthenticated middleware to Passport.
- [x] Add this middleware to the login route.
- [x] Unit tests

## Additional Notes
Related JIRA - https://issues.jboss.org/browse/RAINCATCH-1164

### DIFF
--- a/cloud/passportauth/src/auth/PassportAuth.ts
+++ b/cloud/passportauth/src/auth/PassportAuth.ts
@@ -82,6 +82,7 @@ export class PassportAuth implements EndpointSecurity {
         if (req.session) {
           // Used for redirecting to after a successful login when option successReturnToOrRedirect is defined.
           req.session.returnTo = self.setReturnToUrl(req);
+          req.session.clientURL = req.session.returnTo;
         }
         return res.status(401).send();
       }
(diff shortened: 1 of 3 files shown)


cloud/passportauth/src/auth/PassportAuth.ts
```diff
@@ -107,6 +108,21 @@ export class PassportAuth implements EndpointSecurity {
   }
 
   /**
+   * A middleware which checks if the user is authenticated.
+   * If the user is already authenticated, the user is redirected back to the application,
+   * otherwise it will proceed onto authenticating the user.
+   */
+  public isAuthenticated() {
```
wtrocki: How about wrapping `isAuthenticated` and authService.authenticate into single middleware?

wtrocki (approved):

PR #87, opened 2017-08-18
[RAINCATCH 1169] Reduce data needed from http requests

paolobueno:
See companion PR at https://github.com/feedhenry-raincatcher/raincatcher-angularjs/pull/46

This one makes it so we update the workorder's status field along with the result, this field was being ignored previously AFAIK.


wtrocki:
@paolobueno Do you think it will be hard to integrate results into workorders ?
Is that doable in single ticket?

wtrocki:
@paolobueno 

Bonus question. Do you think it will be hard to integrate results into workorders ?
Is that doable in single ticket?

wtrocki:
PR was reviewed - left comments on angular.js repository PR

wtrocki (approved):

PR #88, opened 2017-08-18
RAINCATCH-1178 hoist dependencies

paolobueno:
Having multiple copies of the same dependencies tends to break typescript compilation, for now hoisting the type declarations solves this type of issue.

Add --hoist as default for the bootstrap command via `lerna.json`.
Update dependencies and converge them to single versions across the packages


lerna.json
```diff
@@ -6,5 +6,10 @@
     "integration/*",
     "demo/*"
   ],
+  "commands": {
+    "bootstrap": {
+      "hoist": "**"
```
wtrocki, line 11: Should we hoist cloud packages only? I'm not sure if we will be able to generate proper shrinkwrap for demo server.

client/datasync-client/package.json
```diff
@@ -33,24 +33,24 @@
     "branches": 50
   },
   "devDependencies": {
-    "@types/chai": "^4.0.1",
-    "@types/lodash": "^4.14.70",
+    "@types/chai": "^4.0.3",
```
wtrocki, line 36: It will be nice to get some github dependency checker to make sure we always stay up to date. 

Did you updated this manually?

wtrocki:
Tested locally. @paolobueno - Let's remove js files from this PR. Looks like we need to have more generic gitignore for js files.

wtrocki:
@paolobueno - It's best to always add steps to verify.
I verified this PR using 
```
lerna clean
lerna bootstrap
```

wtrocki:
@paolobueno - When using this repo with the angular.js and doing bootstrap on angular.js one client packages are installed again.  We will need to disable client packages or hoist angular.js repo as well. Leaving this unmerged for now.

lerna.json
paolobueno: --hoist only works on dev dependencies so it shouldn't affect the regular ones.

If there's anything in devDependencies that's needed for the server to run then it should be a regular dependency.

client/datasync-client/package.json
paolobueno: I started off with `lerna exec -- npm update -S`, expected it to fix all of them to the same one but I still had to address some separate versions manually afterwards.

paolobueno:
@wtrocki Sorry for not including verification steps. I think you need to run:

```
lerna clean
lerna run clean
lerna bootstrap
```

The second one runs the command to delete compiled js files.

wtrocki (approved):

PR #90, opened 2017-08-21
RAINCATCH-914  - api docs generator

wtrocki:
## Motivation

Add script for generating documentation.


wtrocki:
@austincunningham - you can use this api documentation for reference. 
This PR is generator only. Actual PR can be generated once we agreed that changes here are ok.

austincunningham (dismissed):
verified

austincunningham (approved):
verified

PR #92, opened 2017-08-21
[RAINCATCH-1151] Add hashed index to id field for generated collections

paolobueno:
## Motivation
Add indexing on custom `id` field when generating demo data.

## Description
Add extra logging and [hashed](https://docs.mongodb.com/manual/core/index-hashed/) indexing on the id field.

## Progress
- [x] Add index creation before seed data

## Additional Notes
To verify:

- Delete the `raincatcher` mongodb database
- Start the `demo-server` to trigger the seeding of the demo data
- Verify the new index by running `db.collection('workflows').getIndexes()` and `db.collection('workorders').getIndexes()` on a mongo console


wtrocki:
I think that we may even have some separate raincatcher-demo/demo-data package.This will make demo app clean for people wanting framework and sales. leaving up to you if we should do it in this PR

demo/server/src/modules/demo-data/Workorders.ts
```diff
@@ -173,24 +174,39 @@ const workorders = [
   }
 ];
 
+function updateWorkorderDates() {
+  const today = new Date();
+  const tomorrow = new Date();
+  tomorrow.setDate(today.getDate() + 4);
```
wtrocki: That is no longer needed as  I have disabled this fields. They are provided by step when starting

demo/data/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@raincatcher/demo-data",
```
wtrocki, line 2: `@raincatcher-demo/sample-data` - to simplify productization we may need two teams. 
That will help us to differenciate between supported modules and examples.
WDYT?

wtrocki (dismissed):

demo/data/package.json
paolobueno: I think it's totally okay if it helps in that front. For the local dev workflow and lerna nothing would change, only for npm. (and arguably the `node_modules/@raincatcher` + `node_modules/@raincatcher-demo` directory structure, which shouldn't matter)

demo/server/src/modules/index.ts
```diff
@@ -70,7 +71,8 @@ function wfmApiSetup(app: express.Express, connectionPromise: Promise<any>) {
   // Mount api
   const api = new WfmRestApi();
   const role = config.security.adminRole;
-  app.use('/api', securityMiddleware.protect(role), api.createWFMRouter());
+  app.use('/api', securityMiddleware.protect(role));
```
wtrocki, line 74: @JameelB - looks like bug in security middleware prevent from getting proper results. 
This looks like workaround for the moment but we will need to get this tested.

wtrocki (approved):

PR #97, opened 2017-08-24
RAINCATCH-1190 - Extract security interface

wtrocki:


demo/server/src/modules/keycloak/index.ts
```diff
@@ -1,26 +1,38 @@
...
-  // Express Session Configuration.
-  app.use(session(sessionOpts));
-
-  // Create a session store
-  const memoryStore = new session.MemoryStore();
-  const keycloak = new Keycloak({ store: memoryStore }, config.keycloakConfig);
-
-  // Use keycloak middleware & define applications logout route
-  app.use(keycloak.middleware({logout: '/logout'}));
+export class KeycloakSecurity implements EndpointSecurity {
```
wtrocki, line 13: I still think that despite this being really simple we should have it as separate module.

demo/server/src/modules/keycloak/index.ts
```diff
@@ -1,26 +1,38 @@
...
+  private keycloak: any;
+
+  public init(app: express.Express, sessionOpts: session.SessionOptions): void {
+    // Express Session Configuration.
+    app.use(session(sessionOpts));
+
+    // Create a session store
+    const memoryStore = new session.MemoryStore();
+    this.keycloak = new Keycloak({ store: memoryStore }, config.keycloakConfig);
+  }
```
JameelB: also re-add the `app.use(keycloak.middleware({logout: '/logout'}));` to use keycloak middleware and define the logout route

cloud/auth/src/EndpointSecurity.ts
```diff
@@ -0,0 +1,37 @@
...
+  protect(role?: string): express.Handler;
+
+  /**
+   * Create middleware for authentication purposes
+   * This method wraps `passport.authenticate` to provide middleware for authenticating users.
+   *
+   * @param redirect - location to redirect after successful authentication
+   * @param loginError - location to redirect after failed authentication
+   */
+  authenticate(redirect: string, loginError: string): express.Handler;
```
JameelB: would it be possible to remove this from the interface if this only refers to passport?

wtrocki: Everything for you ;P

wtrocki:
@JameelB Really good point about dropping this method. 
It looks like we using that in top level application anyway.

paolobueno (approved):

PR #99, opened 2017-08-28
Update import of EndpointSecurity

JameelB:
## Description
Core fails to compile due to failed import on endpoint security. The endpoint security interface has been moved to express-auth. Update to import EndpointSecurity from express-auth rather than Auth-Passport.


wtrocki (approved):

wtrocki:
Build is failing because of Coverals servers being down. 
Merging as we sending check to coverals as last step and everything else is passing.

PR #101, opened 2017-08-29
An an Example Keycloak Config for the Cloud App

tomjackman:
## Motivation

An an Example Keycloak Config for the Cloud App


wtrocki:
@TommyJ1994 - We cannot merge this change as it is as it will break concept of the config (fields needs to match and be validated with typescript definition. However we will use that example in extracted form and document that properly. Thanks for contribution!

ping @austincunningham

PR #103, opened 2017-08-29
update keycloak prod config

tomjackman:
## Motivation

Update the Keycloak Prod config to use the new OIDC config as per raincatch-1160.

```
"keycloakConfig": {
    "realm": "",
    "bearer-only": true,
    "auth-server-url": "",
    "ssl-required": "",
    "resource": "",
    "use-resource-role-mappings": true
 }
```

### DIFF
--- a/demo/server/config-prod.json
+++ b/demo/server/config-prod.json
@@ -15,9 +15,10 @@
   },
   "keycloakConfig": {
     "realm": "",
+    "bearer-only": true,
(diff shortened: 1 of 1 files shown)


wtrocki:
**Thanks for contributing to Raincatcher**  :1st_place_medal: 
Your change was reviewed and accepted by our team

wtrocki (approved):

PR #104, opened 2017-08-30
DataManger extensions

wtrocki:
## Motivation

Implement forceSync in DataManagers for sync

See: https://github.com/feedhenry-raincatcher/raincatcher-angularjs/pull/64 for mobile changes.

### DIFF
--- a/client/datasync-client/src/DataManager.ts
+++ b/client/datasync-client/src/DataManager.ts
@@ -22,7 +22,7 @@ export class DataManager {
    *
    * @returns Promise (bluebird)
    */
-  public list(): Bluebird<any> {
+  public list(): Bluebird<any[]> {
     const self = this;
     return new Bluebird(function(resolve, reject) {
       syncApi.doList(self.datasetId, function(syncDataSetList: any) {
@@ -164,36 +164,34 @@ export class DataManager {
   /**
    * A utility function to push any updates to the remote server and then stop syncing.
    * If there are pending sync operations, then force them to sync and then stop syncing.
+   * Method can be used to ensure that all changes were saved to server before logout.
+   *
    * @param userOptions - object that can contain delay
    * @returns {*}
    */
   public safeStop = function(userOptions) {
     const self = this;
+    // Amount of time to wait before attempting stop.
     const defaultOptions = {
       delay: 5000
     };
 
     const options = _.defaults(userOptions, defaultOptions);
-
     function forceSyncThenStop(pendingUpdateQueueSize) {
       if (pendingUpdateQueueSize === 0) {
         self.stop().then(Bluebird.resolve);
         return;
       }
+      // Steps: force sync, wait, check if results were synced back, stop server
       return self.forceSync()
         .delay(options.delay)
         .then(self.getQueueSize.bind(self))
         .then(function(size) {
           if (size > 0) {
-            Bluebird.reject(new Error('forceSync failed, outstanding results still present'));
+            return Bluebird.reject(new Error('forceSync failed, outstanding results still present'));
           }
         })
-        .then(self.stop.bind(self))
-        .then(function() {
-          Bluebird.resolve();
-        }, function() {
-          Bluebird.reject(new Error('forceSync error'));
-        });
+        .then(self.stop.bind(self));
     }
     return self.getQueueSize().then(forceSyncThenStop);
   };
@@ -219,7 +217,7 @@ export class DataManager {
     const self = this;
     return new Bluebird(function(resolve) {
       syncApi.getPending(self.datasetId, function(pending?) {
-        resolve(_.size(pending));
+        return resolve(_.size(pending));
       });
     });
   };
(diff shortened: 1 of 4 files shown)


client/datasync-client/src/DataManager.ts
```diff
@@ -113,17 +132,73 @@ export class DataManager {
    *
    * This will stop the sync Client API from sending/recieving updates from the remote server.
    *
-   * @returns {*}
+   * @returns Promise (bluebird)
+   */
+  public stop(): Bluebird<any> {
+    const self = this;
+    return new Bluebird(function(resolve, reject) {
+      syncApi.startSync(self.datasetId, function() {
```
JameelB: this should be stopSync?

client/datasync-client/src/DataManager.ts
```diff
@@ -138,6 +213,18 @@ export class DataManager {
   }
 
   /**
+   * Get the current number of pending sync requests for this data set.
+   */
+  private getQueueSize = function() {
+    const self = this;
+    return new Bluebird(function(resolve) {
+      syncApi.getPending(self.datasetId, function(pending?) {
```
JameelB, line 219: Whats the `?`

wtrocki: short form of `pending?:any`

paolobueno: `pending: any | undefined`

client/datasync-client/src/DataManager.ts
```diff
@@ -18,77 +19,92 @@ export class DataManager {
 
   /**
    * Listing all data for this data set.
+   *
+   * @returns Promise (bluebird)
    */
-  public list(callback: (err?: Error, results?: any) => void) {
+  public list(): Bluebird<any> {
```
paolobueno: I think this could at least be an array type instead of bare `any`, helps client code a little.

client/datasync-client/src/DataManager.ts
```diff
@@ -18,77 +19,92 @@ export class DataManager {
...
-      callback(undefined, dataSetData);
-    }, function handleSyncListError(syncErrorCode: string, syncErrorMessage: string) {
-      callback(new Error(self.formatSyncErrorMessage(syncErrorCode, syncErrorMessage)));
+    return new Bluebird(function(resolve, reject) {
+      syncApi.doList(self.datasetId, function(syncDataSetList: any) {
+        const dataSetData = self.extractDataFromSyncResponse(syncDataSetList);
+        resolve(dataSetData);
+      }, function handleSyncListError(syncErrorCode: string, syncErrorMessage: string) {
+        reject(new Error(self.formatSyncErrorMessage(syncErrorCode, syncErrorMessage)));
+      });
```
paolobueno, line 33: :100:

client/datasync-client/src/DataManager.ts
paolobueno: Is there a way to handle errors for this?

client/datasync-client/src/DataManager.ts
```diff
@@ -113,17 +132,73 @@ export class DataManager {
...
+            Bluebird.reject(new Error('forceSync failed, outstanding results still present'));
+          }
+        })
+        .then(self.stop.bind(self))
+        .then(function() {
+          Bluebird.resolve();
+        }, function() {
+          Bluebird.reject(new Error('forceSync error'));
+        });
+    }
```
paolobueno: Can you add a few code comments to this function? logic's a bit dense

client/datasync-client/src/DataManager.ts
```diff
@@ -113,17 +132,73 @@ export class DataManager {
...
+          if (size > 0) {
+            Bluebird.reject(new Error('forceSync failed, outstanding results still present'));
+          }
+        })
+        .then(self.stop.bind(self))
+        .then(function() {
+          Bluebird.resolve();
+        }, function() {
+          Bluebird.reject(new Error('forceSync error'));
+        });
```
paolobueno: I think this last `then` is redundant, and wrong since it's not `return`ing anything.

If you're just looking to mask the return value and error value, `.then(() => undefined).catch(() => Bluebird.reject(new Error('forceSync error')))` should do the trick.

client/datasync-client/src/DataManager.ts
wtrocki: Actually I was thinking about this as well and aparently this code is just fetching local data. Errors will just return empty object.

client/datasync-client/src/DataManager.ts
wtrocki: To be clear - you mean `any[]` ?

paolobueno: Yes I think that works

client/datasync-client/src/DataManager.ts
paolobueno: :+1: I thought that might be the case, maybe of an error value like -1 or so.

paolobueno (dismissed):

paolobueno (approved):

PR #107, opened 2017-08-31
RAINCATCH-1198 - sync mapper

wtrocki:
## Motivation

Allow to override query filter on server for security reasons. 
Typical usage for filter will be to list results per specific user and do not allow any other filters for security reasons. This fix adds flexible way to do it. 

## Progress
- [x] Enable sync security
- [x] Create mapper
- [x] Add explicit option
- [x] Unit tests
- [x] Documentation

### DIFF
--- a/cloud/datasync/README.md
+++ b/cloud/datasync/README.md
@@ -50,7 +50,7 @@ const router = middleware.createSyncExpressRouter();
 app.use('/', router);
 ```
 
-See [integration](./integration) for complete runnable example.
+See [example app](https://github.com/feedhenry-raincatcher/raincatcher-core/tree/master/cloud/datasync/example) for complete runnable example.
 
 ## Accessing underlying sync library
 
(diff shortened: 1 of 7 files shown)


wtrocki:
###  Verification steps. 

1. Login as Trever.
1. Place breakpoint in custom data handler (mobile)
1. Override assigne
1. Server should still return Trever workorders.

cloud/datasync/README.md
```diff
@@ -66,3 +66,8 @@ Where `anyapicall` is one of the types defined in:
 https://github.com/feedhenry/fh-sync/blob/master/types/fh-sync.d.ts
 
 For more advanced usages please follow [FeedHenry sync documentation](https://github.com/feedhenry/fh-sync/tree/master/docs)
+
+## Overriding sync filter on server
+
+By default sync clients can specify any type of filter and efectivelly get access to
+every record for dataset. If you wish to disable user query_filters for certain datasets for security reasons `SyncMapperMiddleware` should be used, before mounting sync express router.
```
paolobueno: I think this text implies that the mapper middleware is more flexible that just the user data, also the export's name is `userMapperMiddleware`, the SyncMapperMiddleware's filename won't be visible outside the module.

paolobueno (approved):

PR #111, opened 2017-09-01
[Feature] Mobile WFM with single workorder object

wtrocki:
## Motivation

WFM module code was migrated from previous solution. 
Most of the implementation is now spread across 3 main modules (WFM client, workflow, workorder). We need to simplify overal model to use single dataset. 
This will allow us to make other data improvements like having status attached to Workorders and fix some of the major issues in the current solution. 

TL;DR - mobile application should be based only on workorders (single dataset)

UI PR: https://github.com/feedhenry-raincatcher/raincatcher-angularjs/pull/67

## Proposed changes

Workflow results can be merged into workorders and synced together.
Because of their nature workflows doesn't need to be synced. 
They aren't changing often and there is no chance of getting conflict.
Improvements to general step processing and UI framework
Enable duplication in step codes etc. 

## Progress

Work was divided into multiple phases. 

- [x] Disable sync for workflows/results in mobile
- [x] Adjust wfm service to hold data.
- [x] Remove results/workorder services
- [x] Integrate with step definitions change.
- [x] Remove api services and use sync/api directly
- [x] Improve service step index
- [x] Refactor workorder data structure
-  [x] Adding id to steps
-  [x] Introduce version for workflow.
- [x] Adjust UI to not relay on the WorkorderSummary 
- [x] Compute status from workorder without taking results
- [x] Migrate from step code (allow multiple steps with the same code) 
- [x] Remove calls for summary
-  [x]  Using step id instead code.
- [x] Remove summary and other WFM objects that aren't needed.
- [x] Adjust UI step processing to reflect new model
- [x] Include workflow data when creating workoder in portal (Api change) 

## Features that can be included

Add event based processing to monitor and connect to step progression (to be able to control and use branching functionality from code).


wtrocki:
@paolobueno - Posting for early preview. Going to progress this over the weekend to make sure we will have something demoable on monday.

client/wfm/src/model/WorkOrder.ts
```diff
@@ -40,5 +37,16 @@ export interface WorkOrder {
...
+  /**
+   * Contains workflow that will be used for this workorders.
+   * {@link WorkFlow}  this {@link WorkOrder}
+   */
+  workflow: WorkFlow | any;
+
+  /**
+   *
+   */
+  result: WorkOrderResult | any;
```
paolobueno: I think we can remove the `WorkOrderResult` and just have it as `StepResult[]`.

client/wfm/src/model/WorkOrderResult.ts
```diff
@@ -5,9 +5,6 @@ import { StepResult } from './StepResult';
  * On-going progression of a {@link WorkOrder}, contains data for the individual {@link Step}s
  */
 export interface WorkOrderResult {
```
paolobueno: Just get rid of this interface entirely

client/wfm/src/service/WfmService.ts
```diff
@@ -22,6 +22,7 @@ export interface CompleteStepParams {
   stepCode: string;
 }
 
+// FIXME adjust UI to not relay on summary and pick all data from Workorder
```
paolobueno: By the way, apart from the `nextStepIndex` logic, the updated workorder should simply have all the data that was in the summary before.

client/wfm/src/model/WorkOrderResult.ts
wtrocki: Yes - I'm planning to but I'm working different way here. 
I split my work into phases where I do minimal change and then test entire solution.
This phases should be relatively small etc.

client/wfm/src/service/WfmService.ts
wtrocki: nextStepIndex logic will be part of the workoder - Let's cht about it.

client/wfm/src/model/WorkOrderResult.ts
paolobueno: :+1: was just pointing out from the previous comment

wtrocki:
-  [x] Adding id to steps
-  [x] Introduce version for workflow.
-  [x]  Using step id instead code.
- [x] Remove summary and other WFM objects that aren't needed.

paolobueno:
@wtrocki I'll be doing:
- [x] copying over the old wfm entities to the web api to keep them for documentation purposes

wtrocki:
@paolobueno - I cannot do now step code as there is missing index that needs to be fixed first.

paolobueno (approved):

wtrocki:
Verified.

PR #116, opened 2017-09-11
Fixes for npm publishing + platform testing

paolobueno:
## Motivation
Include compiled js files for demo-data and add platform ping endpoint.


wtrocki:
@paolobueno - It looks like core repo should not push client packages as they going to be pushed again in agular. I'm waiting for you to get that cleared.

wtrocki (approved):

PR #119, opened 2017-09-13
[RAINCATCH-1249] Allow to revert to NEW status when backing

paolobueno:
## Motivation
Previously we wouldn't be allowed by the UI to resume a workorder that was backed into the first step. It was considered a UI bug but can also be fixed if we allow workorders to go back into the NEW state.

## Description
- Add branch to return workorder to new state

## Progress
- [X] Go back into NEW status if calling previousStep() on first step
- [x] Unit tests

### DIFF
--- a/client/wfm/src/service/WfmService.ts
+++ b/client/wfm/src/service/WfmService.ts
@@ -92,6 +92,10 @@ export class WfmService {
         // Previous step might be undefined if index-1 < 0
         workorder.currentStep = previousStep && previousStep.id;
       }
+      // if there's no longer a current step, we've backed into NEW status
+      if (!workorder.currentStep) {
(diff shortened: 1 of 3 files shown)


wtrocki:
Really like your way of thinking outside the box. Is state new ok when workorder will have results? Just thinking if that is not going to cause some problems

JameelB:
verified locally

paolobueno:
@wtrocki it's okay to return to `STATUS.NEW` because at least for now going to a previous step deletes the last available result: https://github.com/feedhenry-raincatcher/raincatcher-core/blob/RAINCATCH-1249-workorder-back/client/wfm/src/service/WfmService.ts#L86

JameelB (approved):
Reviewed and verified

PR #122, opened 2017-09-16
RAINCATCH-1262 - server side config improvements

wtrocki:
## Motivation

Application configuration should be centralized and easy to manage using environment variables.

## Description

Improve server side configuration to not relay on the environment variables inside various files in the code. To do that configuration cannot be json anymore. 
We need to have ability to choose from environment variables or default values. 
That's why I have moved configuration to js files. 

JS files will allow us to:
- Add comments to configuration
- Resolve configuration dynamically from environment variables
- Handle legacy parameters (RHMAP)

## Progress
- [x] Move json files to js (for documentation)
- [x] Refactor configuration 
- [x] Add Documentation


demo/server/config-dev.js
```diff
@@ -0,0 +1,77 @@
...
+      // Passport.js secret for JWT tokens. Provide custom value even for development needs
+      "jwtSecret": "D837131FD17F62CB85FBD5919563086369691F4D42379C3596F811839A8992CBA1FBA88DF243BF2481940F112D339C33283BDFEF29A13612550EDAAAB7B5E061",
+      // Allows to provide custom login page messages
+      "portalLoginPage": {
+        "title": "RainCatcher Portal",
+        "invalidMessage": "Invalid Credentials"
+      },
+    },
+    // Keycloak configuration. Uncomment to enable keycloak integration
+    // "keycloak": {
```
JameelB, line 26: Do you think it would be best to fill in the details here? Otherwise we can add a note that they need to fill these in themselves? 

```
   "realm": "raincatcher",
   "bearer-only": true,
   "auth-server-url": "http://localhost:8080/auth",
   "ssl-required": "external",
   "resource": "raincatcher-cloud",
   "use-resource-role-mappings": true
```

JameelB (approved):

demo/server/config-dev.js
wtrocki: hmm.. I think we will cover that in documentation. Missing details means some action. If we provide some values some people may take them for granted and then file bugs that stuff do not work.

PR #125, opened 2017-09-19
[WIP]RAINCATCH-1156 Adding role based check to login

witmicko:
## Motivation
https://issues.jboss.org/browse/RAINCATCH-1156

## Description
Added allowed roles to the server config and cross check them with user roles (intersection of both size > 0) 

## Progress
- [x] Passport
- [ ] Unit tests
- [ ] Documentation

### DIFF
--- a/demo/server/config-dev.js
+++ b/demo/server/config-dev.js
@@ -12,6 +12,7 @@ var config = {
   "security": {
     "adminRole": "admin",
     "userRole": "user",
+    "allowedRoles": ["adminRole"],
     // Passport.js security configuration
     "passportjs": {
       // Passport.js secret for JWT tokens. Provide custom value even for development needs
@@ -19,7 +20,8 @@ var config = {
       // Allows to provide custom login page messages
       "portalLoginPage": {
         "title": "RainCatcher Portal",
-        "invalidMessage": "Invalid Credentials"
+        "invalidMessage": "Invalid Credentials",
+        "insufficientPrivilegesMessage": "Insufficient Privileges"
       },
     },
     // Keycloak configuration. Uncomment to enable keycloak integration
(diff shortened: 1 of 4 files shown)


demo/server/config-dev.js
```diff
@@ -12,14 +12,16 @@ var config = {
   "security": {
     "adminRole": "admin",
     "userRole": "user",
+    "allowedRoles": ["adminRole"],
```
wtrocki, line 15: -1. Roles can be specified directly in the places we need them. 
Why we need this in the server?

witmicko: would you agree to rename this to say `portalAllowedRoles`? I think we should avoid adding this to the client side.

wtrocki:
@witmicko - We cannot have it this way as we provide role based support on top of the client application. We will need to move from admin and user roles fairly soon.

This means that users will need to use role based check - using our AuthService in the portal client to show dialog if user do not have permission as was suggested in the ticket. 
This change needs to work with both keycloak and passworrd. 
Using using interface will allow us to do so.

This aproach works but it's not using parts of our framework which we want others to use. 

ping @JameelB

witmicko:
> This means that users will need to use role based check - using our AuthService in the portal client to show dialog if user do not have permission as was suggested in the ticket.

It does exactly, that. Not sure what is the problem here, pull changes locally and try this.

from what I saw so far passport and keycloak differ a lot in implementation.
We cant have role restriction on the client side - what is stopping me from changing it locally or modifying requests in preflight?

wtrocki:
> from what I saw so far passport and keycloak differ a lot in implementation.

They using single interface that it should be used. If that's not possible then that is a bug as PRD clearly states that this should be possible.

witmicko:
Closing for now in favour of client-side filtering (server side is still protected)

PR #129, opened 2017-09-20
Remove init from interface

wtrocki:
After reviewing code I found that init method it's not used anymore. 
We use different variations of init and that directly.
 

Posting to share opinions.


wtrocki:
@JameelB We going to do that in separate PR that already has some passport changes.

PR #131, opened 2017-09-21
[RAINCATCH-1180] Hide Completed Workorders

paolobueno:
## Motivation
Stop syncing old completed workorders to the mobile client to avoid data clutter and excess data flow.
Keep workorders available for the portal.

## Description
- Add tracking for workorder status
- Add extensibility option for server-side mongodb list filter
- Add filter+configuration for excluding workorders completed X days ago

## Progress
- [X] History tracking
- [X] custom list filters
- [X] completed workorder filter
- [X] Unit tests

### DIFF
--- a/client/wfm/src/model/WorkOrder.ts
+++ b/client/wfm/src/model/WorkOrder.ts
@@ -6,6 +6,12 @@ import { WorkFlow } from './WorkFlow';
 
 type LatLng = [number, number];
 
+interface HistoryEntry {
+  before: STATUS | string;
+  after: STATUS | string;
+  timestamp: number;
+}
+
 /**
  * The executable instance of a {@link WorkFlow}
  */
@@ -69,4 +75,6 @@ export interface WorkOrder {
    * Timestamp of this WorkOrder's last update
    */
   updated: number;
+
+  statusHistory?: HistoryEntry[];
 }
--- a/client/wfm/src/model/WorkOrder.ts
+++ b/client/wfm/src/model/WorkOrder.ts
@@ -6,12 +6,6 @@ import { WorkFlow } from './WorkFlow';
 
 type LatLng = [number, number];
 
-interface HistoryEntry {
-  before: STATUS | string;
-  after: STATUS | string;
-  timestamp: number;
-}
-
 /**
  * The executable instance of a {@link WorkFlow}
  */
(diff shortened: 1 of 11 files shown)


paolobueno:
To verify:

- Complete a workorder via the mobile app
- Manually update the workorder's `statusHistory.Complete` to an older day. i.e. `node -e 'console.log(new Date().setDate(new Date().getDate() - 3))'` in mongo
- Verify that the workorder disappears on the next sync cycle

demo/server/config-prod.js
```diff
@@ -55,7 +55,8 @@ var config = {
   },
   "sync": {
     // Required to handle UI.
-    "customDataHandlers": true
+    "customDataHandlers": true,
+    "excludeOldCompleteWorkOrders": 2
```
wtrocki: Name may be better. Days to filter complete workorders.

demo/server/src/modules/datasync/Connector.ts
```diff
@@ -37,6 +38,10 @@ export function connect() {
       }
       if (config.sync.customDataHandlers) {
         const handler = new GlobalMongoDataHandler(mongo);
+        const excludeDays = appConfig.getConfig().sync.excludeOldCompleteWorkOrders;
+        if (excludeDays > 0) {
+          handler.addListFilterModifier(excludeCompleteWorkOrders(excludeDays));
```
wtrocki: Awesome way of doing that where users can decide if they want this! 
This needs to be documented.

client/wfm/src/model/WorkOrder.ts
```diff
@@ -69,4 +69,17 @@ export interface WorkOrder {
...
+  /**
+   * List of timestamps of when this WorkOrder last reached a given {@link STATUS}
+   *
+   * @example
+   * {
+   *   'New': 1506010024468,
+   *   'Complete': 1506010024468
+   * }
+   */
+  statusHistory?: {
```
wtrocki, line 82: [Question] Do we really need that for this bug?
How about relaying on `status='FINISHED' and updated< moment.daysFromNow(2)`

IMHO after finished workorder is being accepted it should be in read only mode in portal. 
No strong feelings against that, but I will prefer for us to relay on timestamps on results and overall workflow markers to be simple.

demo/server/config-prod.js
wtrocki: Also add comment

client/wfm/src/model/WorkOrder.ts
paolobueno: I actually wanted this to be an array, since we might end up having workorders unassigned and reassigned to other users, they might hit the same status more than once in their lifetime and tracking all of that can lead to more detailed filtering and reporting. But the filter, while doable, could have some performance issues...

I know we don't expect workorders to be updated after they're complete, but there's always stuff like someone in the portal being able to update the workorder name, or maybe we add comments in the future that could also be added after it is complete, so relying on a separate timestamp is more future-proof too.

paolobueno:
@wtrocki addressed comments, thanks for review!

wtrocki (approved):

PR #140, opened 2017-09-30
Add profile filter for passport

wtrocki:
## Motivation

Allow passport framework to filer out user data (including passport) when returning results to client.
screen shot 2017-09-30 at 5 23 39 pm

- [x] Implement check
- [x] Add example
- [x] Add support to demo app
- [x] Testing


JameelB (approved):
Verified locally :+1:

PR #151, opened 2017-10-16
[RAINCATCH-1323] TypeScript-specific start script

paolobueno:
## Motivation
`npm run start` should be the entry point for the compiled js runtime, running the typescript sources via ts-node should have a separate command.

## Description
Added new command to run the demo-server via typescript. Went with `npm run start:ts` since it matches the naming scheme used for the `clean:*` and `publish:*` commands.

## Progress
- [X] Remove extra npm start commands
- [X] Add new command for running via ts-node
- [X] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,9 +160,10 @@ See the example on [examples/js]() for more information also for reusing the uni
 
 - `npm run test` - run unit tests
 - `npm run bootstrap` - perform boostrap for all modules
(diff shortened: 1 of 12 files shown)


wtrocki (approved):
Change looks ok. Add info to release notes for 1.1 that we changed that. Lets also correct getting started guide in docs.

PR #156, opened 2017-10-18
RAINCATCH-1351 - File Server

wtrocki:
## Motivation

WIP work for file server.
Requires some refactoring to fit to typescript model. 

## Progress
- [x] Create interfaces
- [x] Create api mounts
- [x] Create implementations


cloud/file-store/src/FileRoutes.ts
```diff
@@ -0,0 +1,71 @@
...
+    req.fileMeta.id = uuid.create().toString();
+    req.fileMeta.uid = req.fileMeta.id;
+    req.fileMeta.name = req.body.fileName;
+    req.fileMeta.namespace = req.body.namespace;
+    req.fileMeta.owner = req.body.ownerId;
+    req.fileMeta.mimetype = req.file.mimetype;
+    next();
+  };
+
+  router.route('/:filename/binary').post(binaryUploadInitMiddleware, fileService.mutlerMiddleware,
```
paolobueno: Do you think we need these separate endpoints? I think we can use http headers instead.
Seems like multer already only works on top of `multipart/form-data`, so for base64 we could go with `Content-Type: application/octet-stream` and maybe `Content-Transfer-Encoding: base4`, or even `Content-Type: application/base64`

See: https://www.w3.org/Protocols/rfc1341/5_Content-Transfer-Encoding.html

wtrocki: Awesome comment. Fully agree. Going to merge them together!

wtrocki:
@JameelB - I pushed changes to origin by routine :)
New updated version available.

paolobueno (approved):

PR #158, opened 2017-10-19
Added License information to RainCatcher

lgriffin:
## Motivation

Our users have more transparency by us outlining the licensing information that is contained without our dependencies. This will help with adoption of RainCatcher from a legal perspective by making the decision easier.

## Description
I ran the https://github.com/bucharest-gold/license-reporter reporter on the tool and included the output licenses folder. When the license-reporter hits 1.0 stable it will be in NPM, at that point I will send in a 2nd pull request to add it as a dependency to our project allowing it become part of our workflow.

## Progress
- [ ]
- [ ] Unit tests (Not Applicable)
- [ X] Documentation (I have updated the Readme)

## Additional Notes

This will be a requirement for any Red Hat product in the future that might want to consume RainCatcher so this is pre-empting that step while being more transparent to our community.

@wtrocki pinging you for your info

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,3 +174,7 @@ The repository also includes a set of commands for publishing both the standalon
 - `npm run publish:full` - prepares packages and tries to publish to the public registry, requires human input for confirming the version bumps, see the [lerna publish](https://github.com/lerna/lerna#publish) documentation for more information.
 - `npm run publish:dry` - does a dry-run of the publish scripts for local test purposes, avoiding any permanent changes to git history or npm registries
 - `npm run publish:demo-server` - publishes the `demo/server` application to the feedhenry-templates repository as part of the project's inclusion in the [Red Hat Mobile Application Platform](https://www.redhat.com/en/technologies/mobile/application-platform)
+
+### License Information
+
+The licenses folder contains the dependency license information for raincatcher-core. Within the licenses folder you can find a HTML file outlining the dependencies and we helpfully provide a copy of the license file as it appears within the dependency. For convenience, we provide an XML version which can be used with standalone Licenser Tools such as the Node.js [License Reporter](https://github.com/bucharest-gold/license-reporter) project which we used to generate this information.
--- a/licenses/license.html
+++ b/licenses/license.html
@@ -1,0 +1,32 @@
+<html>
+<head>
+<meta http-equiv="Content-Type" content="text/html;charset=utf-8">
+<link rel="stylesheet" type="text/css" href="licenses.css">
+</head>
+<body>
+<h2>raincatcher</h2>
+<table>
+<tr>
+<th>Package Group</th>
+<th>Package Artifact</th>
+<th>Package Version</th>
+<th>Remote Licenses</th>
+<th>Local Licenses</th>
+</tr>
+<tr>
+<td>tslint</td>
+<td>N/A</td>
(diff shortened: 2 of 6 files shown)


lgriffin:
Just to note the content of the folder:

- license.xml -- an XML version of the licenses consumable by other tools
- license.html a simple to view version of the XML for human readability which includes local license links
- license.css some simple styling
- Individual text versions of the dependencies License. This serves two purposes. The first is to show the license info "as is" in the dependency and snapshots it in time, in case the license information changes. Secondly to provide an offline link to the license. In the HTML we provide a remote URL to the various license texts but our users might not have outbound internet connectivity and protects us from firewall rules (popular in insurance companies)

wtrocki:
**Thanks for contributing to Raincatcher**  :1st_place_medal: 

To provide more context for this change.
RainCatcher is being delivered to customers as 3 separate repositories. Each of those are generated from this core repository. Core repository is mono repo that has tools we use as a team to deliver our software. Licenses were run on the top level package. 

*Do we want this change to be included on the tooling repository or the end user deliverable?*

From what I have seen (it's actual requirement from Apache license) we may need to have it on deliverable. 

I have tried to run command it on all packages using:
```
npm i -D license-reporter && lerna exec -- license-reporter
```

This gave me a lots of duplication in license files and many other artifacts.
From what I get the target here will be to get top level view what RainCatcher is using.
I checked where this tool is used and for most of the cases it's being done for user facing source code.

I think that it will be better to:
- Generate licenses on the top level projects that are delivered to customers.
- Have tooling integrated as part of the workflow so we will update things on every release.

We will need to change our release automation to always include this into:
- https://github.com/feedhenry-raincatcher/raincatcher-portal
- https://github.com/feedhenry-raincatcher/raincatcher-mobile
- https://github.com/feedhenry-raincatcher/raincatcher-server

I have created jira ticket to properly facilitate discussion:
https://issues.jboss.org/browse/RAINCATCH-1371

TL;DR - I think we should generate that on different repositories. 
Team will add automation to keep this list up to date.

lgriffin:
Thank you @wtrocki some very sensible comments and observations there.

For clarity, I want to state that there will be duplication by design as the tool has to look at the individual perspective. There will also be a volume of files that are generated which are duplicates by nature. If you have X dependencies with the same license it will produce X local license files with potentially the exact text as it appears. It's a safety net to capture the snapshot from a licensing perspective and more importantly to allow a developer actually see the license text that the original author of our dependency used. Just in case a dependency has an alternative view on the stock text. 

I think personally the licenses should be available on any component that is end user facing in any capacity, so that the developer / customer has a view of what our constituent parts are made up of. It will interestingly provide the licensing information if raincatcher-core is used as a dependency elsewhere -- which is good to show them what we depend on :)

+1 to including this on the automated tooling it means every time we release we snapshot our dependency ecosystem and our licensing. The license-reporter tool in a future version may offer a diff style comparison to flag any changes in text or license type at a dependency level which will provide us with a very welcome layer of security. I'm going to suggest that as a feature enhancement and some cross community collaboration might see it implemented.

wtrocki:
**Thanks for contributing to Raincatcher**  :1st_place_medal: 
Pull request becaome outdated and is going to be closed by our team. 
Feel free to reopen it to provide update or create new one.

PR #159, opened 2017-10-20
WIP file sync client

wtrocki:
## Motivation

Create file sync client. 

## Progress
- [x] Create module with queue implementation


client/filestore-client/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "version": "1.0.0",
+  "description": "RainCatcher file storage client",
+  "types": "src/index.ts",
+  "author": "feedhenry-raincatcher@redhat.com",
+  "license": "Apache-2.0",
+  "main": "src/",
+  "scripts": {
+    "clean": "del coverage_report src/**/*.js src/**/*.map test/**/*.js test/**/*.map",
+    "build": "tsc",
+    "start": "ts-node src/index.ts",
```
paolobueno: No need for a `start` command on this package, maybe we should remove it from the base?

client/filestore-client/src/CordovaFileUpload.ts
```diff
@@ -0,0 +1,62 @@
+import * as Bluebird from 'bluebird';
+
+// Cordova mocks for typescript
+declare var FileUploadOptions: any;
+declare var FileTransfer: any;
```
paolobueno: Maybe put these in private fields to a `CordovaFileUpload` class, or just rename the file so it doesn't suggest it's exporting a class.

client/filestore-client/src/CordovaFileUpload.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @param fileURI
+ * @param options
+ * @returns {*}
+ */
+export function uploadFile(baseUrl, userId, fileURI, options) {
+  if (arguments.length < 2) {
+    return Bluebird.reject('userId and fileURI parameters are required.');
+  } else {
+    options = options || {};
+    const fileUploadOptions = new FileUploadOptions();
```
paolobueno: Can be just an interface instead of a class

client/filestore-client/src/CordovaFileUpload.ts
wtrocki: I had problem with this. There are definitions for this but they are outdated (more than year old) when bringing old definitions I could have some differences between the new ones
with such small exposure and couple trials I decided to go this way. 

I will wrap methods and fields into class.

client/filestore-client/src/CordovaFileUpload.ts
```diff
@@ -0,0 +1,62 @@
...
+    fileUploadOptions.fileKey = options.fileKey || 'binaryfile';
+    fileUploadOptions.fileName = options.fileName;
+    fileUploadOptions.mimeType = options.mimeType || 'image/jpeg';
+    fileUploadOptions.params = {
+      ownerId: userId,
+      fileName: options.fileName
+    };
+    const timeout = options.timeout || 2000;
+    const retries = options.retries || 1;
+    const serverURI = baseUrl + '/upload/binary';
```
wtrocki: This is something I will need to extract and improve now. 
Actual network client needs to come from developers as they need to apply some headers for security etc.

@paolobueno - Do you think we can put that into angularjs-http module? Thinking how to get that without creating additional artifacts.

client/filestore-client/src/FileManager.ts
```diff
@@ -0,0 +1,82 @@
...
+      self.startProcessingUploads();
+    });
+  }
+
+  public startProcessingUploads() {
+    const queueItems = this.uploadQueue.restoreData().getItemList();
+    const self = this;
+    if (queueItems && queueItems.length > 0) {
+      console.info('Processing offline file queue. Number of items to save: ', queueItems.length);
+      let promise = this.saveFile(queueItems[0]).then(function() {
```
paolobueno: The queue processing part might be the hardest part of the file feature...

I think the best function to deal with the concurrency for now would be
```typescript
Bluebird.map(queueItems, i => self.saveFile(i), {concurrency: 1});
```

See http://bluebirdjs.com/docs/api/promise.map.html#map-option-concurrency

client/filestore-client/src/FileManager.ts
```diff
@@ -0,0 +1,82 @@
...
+    });
+  }
+
+  /**
+   * Add file to upload queue. File would be uploaded depending on internet connectivity.
+   *
+   * @param file {userId, fileURI, options, dataUrl}
+   * @returns {*}
+   */
+  public scheduleFileToBeUploaded(file: any) {
```
paolobueno: TODO: add interface for the file param obj

client/filestore-client/src/FileQueue.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Queue implementation backed by browser persistent storage
+ *
+ * @param name
+ */
+export class FileQueue {
+  private queueName: string;
+  // TODO interface for datatype
+  private queueData: any[];
+
+  public constructor(name: string) {
```
paolobueno: We might need to add `localStorage` as a constructor DI for unit tests.

client/filestore-client/src/CordovaFileUpload.ts
paolobueno: Ah, you mean ts definitions for the cordova plugins? Sorry, didn't know these were from cordova, thought they were new classes.

I was referring to the actual name of the file though, the module is exporting the individual functions, not a class, so I think we should avoid naming it starting with uppercase.

client/filestore-client/src/CordovaFileUpload.ts
```diff
@@ -0,0 +1,62 @@
...
+
+// Cordova mocks for typescript
+declare var FileUploadOptions: any;
+declare var FileTransfer: any;
+
+// Handling file upload to server
+function fileUpload(fileURI, serverURI, fileUploadOptions) {
+  const transfer = new FileTransfer();
+  return new Bluebird(function(resolve, reject) {
+    transfer.upload(fileURI, serverURI, function(response) {
```
wtrocki: @JameelB This will fail currently as file transfer needs to work with passport/keycloak. 


We can pass headers here but this will mean that there will be 2 different places to edit those.
See: http://ngcordova.com/docs/plugins/fileTransfer/

@JameelB - What we should do here in your opinion:

Should we move this to the top level app to have control over the the options and allow users to control this?

client/filestore-client/src/CordovaFileUpload.ts
wtrocki: They are globals from cordova libraries required in top level app. 
Sadly cordova requrires them to be on the top level application. as they are part of the build process.

TL;DR - what is your suggestion here :)?

client/filestore-client/src/CordovaFileUpload.ts
paolobueno: Makes sense to reutilize our own angularjs-http module. I think this module will need to receive a 'compatible' http client as a dependency and defer to it for the actual communications.

I've taken a look at the file transfer plugin and I think it's not required for this, and we can use any http lib instead as long as we can configure the correct mime types and headers: https://github.com/apache/cordova-plugin-file-transfer#deprecated

We probably need to add support for file uploads to the angularjs-http module, since it was built only as a json-based REST client. This can be done like this: https://stackoverflow.com/a/25264008/2600345, maybe adding something like `HttpDataService.prototype.uploadFile(relativeUrl, fileData)`

What I'm worried about right now is introducing a circular dependency between both core and angularjs repositories.

paolobueno: A way to avoid coupling this module with `angularjs-http` would be for it to only receive a simple callback `uploadFile(url, data) => Promise` and let the demo application/step configure it to come from the angularjs module.

client/filestore-client/src/CordovaFileUpload.ts
paolobueno: Some discussion around this topic in https://github.com/feedhenry-raincatcher/raincatcher-core/pull/159#discussion_r146268198.

tl;dr, we could ditch the cordova file transfer plugin and just have this module receive a function that'll handle the file upload itself.

client/filestore-client/src/CordovaFileUpload.ts
wtrocki: > What I'm worried about right now is introducing a circular dependency between both core and angularjs repositories.

Let's bind them in the reference app so it will not be hard dependency.

client/filestore-client/src/FileManager.ts
wtrocki: This one definitely need improvements. Great point!

client/filestore-client/src/CordovaFileUpload.ts
wtrocki: @paolobueno That will be the best aproach as we can put that on http module or even use file transfer directly in the application!

JameelB: @wtrocki @paolobueno if its possible to use angular-http i vote for doing it that way

client/filestore-client/src/CordovaFileUpload.ts
wtrocki: I do not understand what you mean here.

client/filestore-client/src/CordovaFileUpload.ts
paolobueno: So even if keeping the filetransfer plugin, what I found in other `@types/cordova-plugin-*` definitions is that they add the objects under `window.*` or `window.navigator.*`, and any browser global can be referenced like that too.

client/filestore-client/src/CordovaFileUpload.ts
paolobueno: Yeah, disregard this, didn't knew it came from cordova

PR #160, opened 2017-10-23
Fix compilation error

JameelB:
## Description
Core fails to start due to the following typescript compilation error:
`Unable to compile TypeScript: ../../cloud/wfm-rest-api/src/WfmRestApi.ts (21,19): Expected at least 1 arguments, but got 2.`


wtrocki:
@JameelB - I couldn't reproduce this problem, but got your changes and it works as well.. 
What I think is that there is specific version of the definition that is broken. 
I think it will be best to drop that as it gives no value.

wtrocki (approved):

paolobueno (approved):
Not sure how this fixes a lodash-related compilation error either, but okay to merge.

PR #164, opened 2017-10-25
[RAINCATCH-1350] Add camera package to support camera step

paolobueno:
## Motivation
Move camera class from spike to core, use the file upload client from it

## Progress
- [X] Add Camera package
- [X] Link filestorage-client
- [X] Use upload queue to upload taken photos (untested)
- [ ] Use download queue
- [ ] Documentation


client/camera/src/Camera.ts
```diff
@@ -0,0 +1,70 @@
...
+        const blob = base64ToBlob(uri, 'image/jpg');
+        resolve(window.URL.createObjectURL(blob));
+      }))
+      ).then(function(uri) {
+        const file: FileEntry = {
+          uri,
+          userId: "max"
+        };
+        // Fire and forget upload for now
+        // TODO: figure out how to report uploads to UI
```
wtrocki, line 65: We need 2 notification streams:
- One for the files that were uploaded (to mark that they were synced back to server)
- One for files that should be downloaded. We do not have this case in our demo step - 
Maybe we should.

client/filestore-client/src/FileManager.ts
```diff
@@ -93,18 +97,11 @@ export class FileManager {
    * Upload file to server.
    * Function would choose right method depending on parameters.
    *
-   *  // FIXME verify if we need to handle both cases.
    * @param file {fileURI, dataUrl}
    * @returns {*}
    */
-  private createFile(file) {
```
wtrocki, line 100: ❤️

wtrocki:
@paolobueno We can do download queue as part of separate ticket. Merging this back to feature branch to make cross module changes.

PR #165, opened 2017-10-26
File storage support feature

wtrocki:
## Motivation

See: https://issues.jboss.org/browse/RAINCATCH-1349


austincunningham (dismissed):
don't merge

wtrocki:
Snyk build is failing due to outdated published module. 
Local build for snyk is working fine.

wtrocki:
It looks like that we still getting error:
```

Argument of type 'typeof rename' is not assignable to parameter of type '(arg1: string, arg2: string, callback: (err: any, result: void) => void) => void'.
  Types of parameters 'callback' and 'callback' are incompatible.
```

austincunningham (approved):

wtrocki:
@paolobueno Can I just rebase and merge this? It's required to fix angularjs build.

wtrocki:
@paolobueno I looks like we sending back the local uri to server. Curious if that is needed? 

screen shot 2017-11-20 at 12 13 49 pm

I noticed that download (sync changes from the server) is not included in the demo. We demoing i only for the portal (using different http api). I think that it's ok to go without this option, but worth to mention that we support sync from server. Is that ok?

Going to test it to make sure it works.

wtrocki:
Download works fine! I couldn't really figure out how to replicate it in the standard step so we going to leave that without example.

PR #169, opened 2017-10-26
Fix types error for api-controller and demoDataSetup

JameelB:
## Description
Fix Typescript types errors

## Additional Notes
Related Gist - https://gist.github.com/JameelB/bca467389fd1b58ea027b36b625e9793

### DIFF
--- a/demo/server/src/modules/index.ts
+++ b/demo/server/src/modules/index.ts
@@ -88,16 +88,24 @@ function wfmApiSetup(app: express.Router, connectionPromise: Promise<any>) {
   const role = config.security.adminRole;
   app.use('/api', portalsecurityMiddleware.protect(role));
   app.use('/api', api.createWFMRouter());
-  connectionPromise.then(function(mongo: Db) {
-    // Fix compilation problem with different version of Db.
-    api.setDb(mongo as any);
-  });
+  if (!connectionPromise) {
+    getLogger().error('Failed to connect to a database');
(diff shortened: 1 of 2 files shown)


demo/server/src/modules/index.ts
```diff
@@ -94,7 +94,7 @@ function wfmApiSetup(app: express.Router, connectionPromise: Promise<any>) {
   });
 }
 
-function demoDataSetup(connectionPromise: Promise<Db>) {
+function demoDataSetup(connectionPromise: Promise<any>) {
```
paolobueno: If the promise that's being passed here is really `Db | undefined` from line 31 I think we should look into where the undefined is coming from, might introduce a bug.

paolobueno (dismissed):
Great fix, just a little bit worried about the mongodb being possibly undefined.

demo/server/src/modules/index.ts
JameelB: That's due to the function that's giving back the connectionPromise as that may be void. https://github.com/feedhenry-raincatcher/raincatcher-core/blob/master/demo/server/src/modules/index.ts#L66-L83

paolobueno: Yes, might be worth it to handle that case in this function, even if it's just erroing out and printing it to the user.

wtrocki (approved):

PR #174, opened 2017-10-27
fix file-store api

JameelB:
## Description
Fixed and added securityMiddleware to the filestore API\

NOTE: Endpoint for filestore has been changed to `/file/binary/` or `/file/base64/`

### DIFF
--- a/cloud/filestore/src/FileRoutes.ts
+++ b/cloud/filestore/src/FileRoutes.ts
@@ -11,8 +11,9 @@ import * as fileService from './services/FileService';
  * @returns router instance of express router
  */
 export function createRouter(storageEngine: FileStorage) {
+  fileService.createTemporaryStorageFolder();
   const router = Router();
-  router.route('/:filename').post(function(req, res, next) {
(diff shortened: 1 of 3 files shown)


wtrocki (approved):
Verified. Works like dream!

wtrocki:
@JameelB Going to adjust client to new API.

cloud/filestore/src/FileRoutes.ts
```diff
@@ -11,8 +11,9 @@ import * as fileService from './services/FileService';
  * @returns router instance of express router
  */
 export function createRouter(storageEngine: FileStorage) {
+  fileService.createTemporaryStorageFolder();
   const router = Router();
-  router.route('/:filename').post(function(req, res, next) {
+  router.route('/base64/:filename').post(function(req, res, next) {
```
paolobueno, line 16: Do you guys know the exact format needed for `parseBase64Stream`? Guess we can just test it out once this is merged :+1:

wtrocki, line 16: Looks like filename is not used/needed. We can remove it in separate PR.

PR #175, opened 2017-10-31
Add extra metadata to post request

paolobueno:
ping @JameelB

### DIFF
--- a/client/filestore-client/src/CordovaFileSupport.ts
+++ b/client/filestore-client/src/CordovaFileSupport.ts
@@ -1,4 +1,5 @@
 import * as Bluebird from 'bluebird';
+import { each } from 'lodash';
 import { FileQueueEntry } from './FileQueueEntry';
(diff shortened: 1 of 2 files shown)


client/filestore-client/src/CordovaFileSupport.ts
```diff
@@ -43,6 +44,11 @@ export class CordovaFileSupport {
           const blob = new Blob([new Uint8Array(this.result)], { type: 'image/jpg' });
           const data = new FormData();
           data.append('file', blob);
+          each(file, function(value, key) {
+            if (value) {
+              data.append(key, value);
+            }
+          });
```
paolobueno, line 51: This just adds extra fields from the FileQueueEntry object to the request, currently the server ignores them, but I think it's good to send them too for future-proofing (TM)

JameelB (approved):

PR #176, opened 2017-11-01
[RAINCATCH-1389] Support base64 upload

paolobueno:
## Motivation
Add support for running the gallery step in a browser by being able to upload the base64-encoded images.

## Description

- [X] Convert base64 data to Blob
- [X] Support upload of base64-converted binary data
- [x] Remove unused `base64/` endpoints

### DIFF
--- a/client/camera/src/Camera.ts
+++ b/client/camera/src/Camera.ts
@@ -7,24 +7,19 @@ import { buildCameraOptions } from './buildCameraOptions';
 
 declare var window;
 
-export const TYPE_IMAGE_BASE64 = 'BASE64';
-export const TYPE_IMAGE_URI = 'URI';
-
 /**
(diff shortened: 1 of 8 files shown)


client/camera/src/Camera.ts
```diff
@@ -48,6 +43,8 @@ export class Camera {
         const userOptions = optionsFn(navigator.camera);
         options = _.merge(options, userOptions);
       }
+      // assign options object to self for convenience
+      this.options = options;
```
JameelB, line 47: would you need self here?

JameelB (approved):

client/camera/src/Camera.ts
paolobueno: No, because of the arrow functions :D

PR #187, opened 2017-11-20
Do not fail when camera is undefined

wtrocki:
## Motivation

See: https://github.com/feedhenry-raincatcher/raincatcher-angularjs/pull/130


paolobueno (approved):
There's a similar guard on `Camera#init` but this is good to have too :+1:

PR #191, opened 2017-11-23
Documentation updates

wtrocki:
## Motivation
 
Review and supplement existing file documentation.

### DIFF
--- a/cloud/filestore/example/index.ts
+++ b/cloud/filestore/example/index.ts
@@ -1,8 +1,6 @@
 import { getLogger } from '@raincatcher/logger';
 import { createRouter, FileMetadata, FileStorage, LocalStorage } from '../src/index';
 
-const logger = getLogger();
-
 // Create express middleware
 import * as bodyParser from 'body-parser';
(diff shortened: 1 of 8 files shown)


cloud/filestore/example/index.ts
```diff
@@ -20,7 +18,7 @@ const router = createRouter(engine);
 app.use('/', router);
 
 app.listen(3000, function() {
-  logger.info('Example app listening on port 3000!');
+  console.info('Example app listening on port 3000!');
```
austincunningham: Why change to console.info ?

cloud/filestore/example/index.ts
```diff
@@ -1,8 +1,6 @@
 import { getLogger } from '@raincatcher/logger';
```
austincunningham: if removing logger remove this as well

cloud/filestore/example/index.ts
wtrocki: Going to restore it.

paolobueno (approved):

PR #193, opened 2017-11-24
bump docs api version in docgen.sh

austincunningham:
## Motivation
bump docs api version in docgen.sh as publishing docs breaks api docs links 

## Description
bump version no.
## Progress
- [ x ] Documentation

### DIFF
--- a/scripts/docgen.sh
+++ b/scripts/docgen.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 ## Script for generating core modules documentation
 
-VERSION=1.0.0
+VERSION=1.1.0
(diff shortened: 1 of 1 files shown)


paolobueno (approved):

paolobueno:
We were using this script yesterday and had some trouble, for now I had to manually copy over the generated typedocs to the right spot, created an issue to fix this: https://issues.jboss.org/browse/RAINCATCH-1399

So while this change is necessary I'm not sure the script will work right now :sweat_smile:

paolobueno:
BTW ideally this should pick up the version number in `/lerna.json` in the future

PR #194, opened 2017-11-24
WIP [RAINCATCH-1372] Update TSC and other packages

paolobueno:
## Motivation
Update typescript compiler version and other package versions.

Also fix documentation script bugs. (RAINCATCH-1399)

## Progress
- [x] Update tsc
- [x] Update other packages
- [x] Removing hoisting requirement from core
- [x] Create base tsconfig on repository root


client/datasync-client/tsconfig.json
```diff
@@ -1,14 +1,7 @@
...
-    "noImplicitAny": false,
-    "experimentalDecorators": true,
-    "strictNullChecks": true,
-    "sourceMap": true
-  },
-  "include": [
-    "src/",
-    "test/"
-  ]
+    "extends": "../../tsconfig.json",
```
wtrocki, line 2: Great idea!

wtrocki:
@feedhenry-raincatcher/rcdev  After this change `npm install` will be required (to instal new learna) and complete bootstrap. IT will also require new version of lerna.

wtrocki:
@paolobueno - Did that compile to you? Getting multiple problems like this.
```
../../cloud/passportauth/src/auth/PassportAuth.ts (39,15): Argument of type 'RequestHandler' is not assignable to parameter of type 'PathParams'.\n  Type 'RequestHandler' is not assignable to type '(string | RegExp)[]'.\n    Property 'push' is missing in type 'RequestHandler'.

``

File: client/camera/package.json [hunk 2]
```diff
@@ -34,28 +34,28 @@
     "branches": 80
   },
   "devDependencies": {
-    "@types/bluebird": "^3.5.8",
-    "@types/chai": "^4.0.3",
+    "@types/bluebird": "^3.5.18",
```
wtrocki: I think that using `^` is the main source of all our compilation problems. 
Not locking types brings some incompatibilities across the board. If we do not want to go back to shrinkwrap best to lock them - possibly separate ticket.

paolobueno: It seems `npm-check-updates` has a --removeRange flag, I'll give it a try.

paolobueno: Ideally we could start including the new npm 5+ package-locks in version control too.

paolobueno:
I'm not getting any compilation errors, I did have some in travis that I fixed via removing a bad @types declaration, maybe a `lerna clean` would fix that one for you. Thanks for testing!

wtrocki (dismissed):
Fixed by lerna clean

wtrocki:
@paolobueno I have done lerna clean (it took a while to install all stuff again) but I can confirm that it works. I just forgot to update PR and approve this change.

wtrocki:
@paolobueno e2c83cd updates only wfm module.

wtrocki:
Changes tested with clean install.

wtrocki (approved):

PR #196, opened 2017-12-01
Raincatch 1402

wtrocki:
## Motivation

Hot-fix for windows path issues

### DIFF
--- a/cloud/datasync/src/web/SyncWebExpress.ts
+++ b/cloud/datasync/src/web/SyncWebExpress.ts
@@ -5,7 +5,6 @@
 import { getLogger } from '@raincatcher/logger';
 import * as Express from 'express';
 import * as sync from 'fh-sync';
-import * as path from 'path';
 const logger = getLogger();
 /**
  * Expose Feedhenry Sync API using express middleware
@@ -27,7 +26,7 @@ export class SyncExpressMiddleware {
    * Create express router for sync endpoints
    */
   public createSyncExpressRouter() {
-    const apiURI = path.join(this.prefix + '/:datasetId');
+    const apiURI = this.prefix + '/:datasetId';
     logger.debug('Creating sync endpoint');
     const syncRoute = this.router.route(apiURI);
 


austincunningham (dismissed):

cloud/datasync/src/web/SyncWebExpress.ts
```diff
@@ -23,7 +22,7 @@ export class SyncExpressMiddleware {
    * Create express router for sync endpoints
    */
   public createSyncExpressRouter() {
-    const apiURI = path.join(this.prefix + '/:datasetId');
+    const apiURI = this.prefix + '/:datasetId';
```
paolobueno, line 29: :+1: so it seems that the right method URLs would be [`url.resolve`](https://nodejs.org/api/url.html#url_url_resolve_from_to) to avoid problems in windows.

I've did a quick search through the codebase and I think this is the only instance where this method is being used with urls.

paolobueno (dismissed):

cloud/datasync/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@raincatcher/datasync-cloud",
-  "version": "1.1.0",
+  "version": "1.1.1",
```
paolobueno: We should defer the version bump to `lerna publish`

cloud/datasync/src/web/SyncWebExpress.ts
wtrocki: I copied over example with url resolve, but it's not valid for this use case: 
```
url.resolve('/one/two/three', 'four');         // '/one/two/four'
```

Simple concatenation will do the job here.

paolobueno: Yeah, considering that we were already passing the `+` concatenated value to `path.join` it wasn't even doing anything before except wrongly replacing the directory separator in windows.

paolobueno (approved):

PR #198, opened 2017-12-07
[RAINCATCH-1409] Move types to dependencies and check dependency usage

paolobueno:
## Motivation
Add @types to dependencies to allow usage from other typescript projects and check whether we're actually using some of the dependencies.

## Description
Add depcheck command
Check dependencies on the packages
Move types dependencies to regular deps from devDeps


wtrocki:
I understand the rationale for this change. Really cool to have dependency checker.
tried that with complete clean and it works fine. 

Good read: https://github.com/Microsoft/types-publisher/issues/81

@paolobueno - I think it's worth to do release to verify new behavior. WDYT?

wtrocki (approved):

paolobueno:
@wtrocki In relation to verifying the behavior I think it might be nice to do a new patch release, but if that's so I'd like to defer merging this PR until I do the same dependencies revision on the angularjs repo.

Though there's no @types issue there, we might be able to reduce bundle sizes very nicely by reviewing our dependencies.

PR #199, opened 2017-12-12
Readd browserify-shim since it's a hidden dependency of fh-sync-js

paolobueno:
## Motivation
fh-sync-js fails to build without the hidden peerDependency of browserify-shim


wtrocki:
That's common issue that will be resolved in the sync. 
Temporary fix in our repo will help a lot.

wtrocki (approved):

